PR #26, opened 2015-03-27
Fix variable declarations triggering some reference errors

Ulflander:
Some variables with bad declaration syntax in indefinite_article.js caused some reference error, so this function was not usable:
```
nlp.noun('names').article()
```
This pull request fixes it.

Cheers!


spencermountain:
thanks!

PR #27, opened 2015-04-07
Fix behaviour under prototype pollution

gunyarakun:
If prototype of Object(or Array) is polluted, nlp_compromise doesn't work well because of 'in' operator.
I fixed them.


spencermountain:
thanks Tasuku! very nice!

PR #30, opened 2015-04-28
Added common "latin" abbreviations ...

redaktor:
e.g. and i.e. as 
```js
["...", "e.g", "i.e"]
```
See http://www.elearnenglishlanguage.com/blog/english-mistakes/eg-vs-ie/ saying
>"The Latin abbreviations e.g. and i.e. are commonly used in English ..." (?)

Sentences might split due to punctuation in abbreviations if those are not listed...


spencermountain:
thanks Sebastian, that's a good find.

PR #55, opened 2015-07-27
Correct nlp.ngram options parameter name

cyanCYMK:
Added clarification that `options.max_size` is the maximum gram count. Corrected `max_gram` to `max_size` as there is no `max_gram` option for that method.


spencermountain:
thanks!

PR #75, opened 2016-01-02
Add main entry

aboutlo:
Currently to use `nlp_compromise` in a node style env (e.g. webpack, browserify, node) it's required specify the full path. 

e.g. 
```
import nlp from "nlp_compromise/src/index"
```

Add the `main` entry in the package.json allow to use just
so that we can just do

```
import nlp from "nlp_compromise"
```


spencermountain:
ah thanks!
i'll add this change too to the 2.0 branch.
cheers!

PR #82, opened 2016-02-01
Removed duplicate declaration

creatorrr:
Strict mode throws a tantrum for the duplicate declaration:

```SyntaxError: Duplicate data property in object literal not allowed in strict mode```

Removed the duplicate declaration to avoid build errors in tools like babel etc.


spencermountain:
ah! thanks Diwank!

Issue #84, opened 2016-02-04
Pass the library instance to the plugin


creatorrr:
Hey!

So, one of the use cases that popped up is that some plugins might need access to the library instance. A random example, let's say we want to return the adjective form of an adverb as an instance of the `Adjective` class:

```javascript

let Mixin = {
  Adjective: {
    to_adverb_instance: function () {
      return NLP.adjective(this.to_adjective());
    }
  }
}

```

There doesn't seem to be any way of doing this at the moment other than requiring the library itself but I think that'd be an unnecessary overhead for a mixin.

Maybe, an alternate definition like this could work?

```javascript

let Mixin = NLP => {
  return {
  Adjective: {
    to_adjective_instance: function () {
      return NLP.adjective(this.to_adjective());
    }
  }
  }
}

```

spencermountain:
ah! i like this a lot.
I'm glad you've been able to follow the mixin idea. I wanna make it great, and I'm happy to change it.

I agree that having access to the library itself makes a lot of sense, and that's a great example.

so in this version, the mixin would require NLP itself, and augment it on its own? I mean, to use it would be:
```javascript
nlp=require("nlp_compromise")
require("that_cool_mixin")
nlp.adjective.that_cool_mixin()
```
am I right?
.. if we had more than one mixin, would it still work?

spencermountain:
(this is really not my expertise, so I'm very keen to have some help organising this) So thanks in advance

creatorrr:
Hmm, I don't think this particular way would solve the problem unfortunately. Because any implementation of such a plugin would either have to depend on a global var or require the library internally. :-/

I like the current form quite a bit as it makes it explicit that you are mutating the library interface. I was thinking more along the lines that the `NLP.mixin` function behave as it is (for backwards compatibility as well) but it also accepts a function instead, that when passed an instance of the library returns an object similar to the mixin object. For instance,

```javascript
// The new mixin function could accept the following
let NLP = require("nlp_compromise");

NLP.mixin( library => {
  return {
    Term: {
      hola: () => console.log("hola")
    }
  };
});
```

What do you think? If you like this idea, I'm happy to make a pull request. :)

spencermountain:
jeez that's clever. love it.
please do it

creatorrr:
Done #91

PR #91, opened 2016-02-06
Change Plugin API to pass library instance to plugins

creatorrr:
Changed the plugin system. Updated the docs. Left out the following (so you can review):
 - README, CHANGELOG update
 - Also, `npm test` and `grunt docs` were failing on my machine. Is it just me or do I need to fix that?

Closing #84


spencermountain:
YES!

PR #112, opened 2016-05-11
Proposed fix for nlp.text().sentences() for quoted strings containing punctuation.

TransWebT:
In attempting to parse a sentence containing something like: 'my double quoted "string!".' I found that the sentence_parser() function would break the string prior to the ending double quote.  The included change fixes this case - I'd like to know if it causes any unexpected side effects.
 
Also, added a demonstrator for nlp.text().sentences() to demo/index.html.  Typing multiple sentences into the new textarea field and pressing the Parse button will parse the typed paragraph into sentences.


spencermountain:
beauty!

spencermountain:
yeah, that makes sense. Thanks for your work!

TransWebT:
you're welcome - thank you for the great API!

PR #123, opened 2016-05-14
document the case of names that do not signal gender

vielmetti:
As noted in #117 the software attempts to guess the gender of names. The internal and external documentation should show that three gender values are defined: 'Male', 'Female', and null.

### DIFF
--- a/src/term/noun/person/gender.js
+++ b/src/term/noun/person/gender.js
@@ -38,9 +38,11 @@ const gender = function(normal) {
   if (firstName.match(/(nn|ll|tt)/i)) { //if it has double-consonants, female
     return 'Female';
   }
+  // name not recognized, or recognized as of indeterminate gender
   return null;
 };
 module.exports = gender;
 
 // console.log(gender('john', 'john') === 'Male');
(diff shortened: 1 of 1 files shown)


seandougall:
And in the interest of accuracy, `return null;` should be moved to the first line of the function body, because no name in existence denotes gender unambiguously.

RRorg:
You chose a bad example for `null`, since `Jan` is a valid first name (a form of Johannes in many European countries), and the return value is simply highlighting a shortcoming of the library. I recommend using a completely made-up "name".

vielmetti:
Jan is an American female name, short form for Janet. The example illustrates that the library does not distinguish currently between unknown names of unknown gender vs. known names of indeterminate gender.

-----Original Message-----

spencermountain:
thanks edward, will merge this later in the week, when things quiet down. ;)
thanks for your help.

alexbrault:
@RRorg: returning null doesn't mean "this is not a valid name", it means "I can't tell whether it refers to a male or female individual".

RRorg:
Maybe working with confidence scores instead would be a better solution.

PR #124, opened 2016-05-15
Fix normalization section

jacobmischka:
Need newlines to fix headings, fix incorrect function call, add missing parens.


spencermountain:
ah, thanks jacob.

PR #136, opened 2016-05-17
few new dev-deps, tasks for running demo, uses npm scripts

khtdr:
Thanks for sharing your work 👍 

My commits are small and the comments are specific. This may make it easier for others who, like me, don't have grunt or browserify installed globally.


spencermountain:
ah! very clever.
This is really helpful. Thanks for this

PR #137, opened 2016-05-17
Fix sentence.replace() which fails on not found terms

ptmt:
Minor fix to protect from `TypeError: Cannot read property 'forEach' of null`.

Reproducing test included:

```js
['the boy and the girl', 'the cat', 'the house', 'the boy and the girl']
```


spencermountain:
thanks Dima! The work you're doing with r-native is incredible.
cheers!

PR #140, opened 2016-05-17
Add JPY as a currency.

Irishsmurf:
The ISO code JPY is used pretty often when referring to the Japanese Yen.
Added that as a Currency Unit.

I've also moved GBP down towards the other ISO symbols.


spencermountain:
thanks David!

Issue #129, opened 2016-05-15
New Unit: Grams


2426021684:
Parsing "5 grams" results in no unit but parsing "5 g" works.

spencermountain:
thanks! will take a look

PR #142, opened 2016-05-17
Support for Plural Units

zeepeeare:
Hey, this is a proposed fix for #129 , with some added tests to demonstrate the behavior.

Thanks


spencermountain:
ah!! oopsies.
thanks man!

PR #144, opened 2016-05-17
added year regex to is_date

andrewsalveson:
years, by themselves as numbers, were not showing up as dates


spencermountain:
thanks Andrew! I may add a check to ensure it's under 3000 or something
thanks!

PR #146, opened 2016-05-18
Fix prefixes of metric units of measurement & add (and correct some) demonyms

ilyankou:
According to NIST (and WolframAlpha, and Wikipedia)
http://www.nist.gov/pml/wmd/metric/prefixes.cfm


ilyankou:
Add demonyms

spencermountain:
YAHOO! :muscle:

PR #149, opened 2016-05-19
Add HUF as a currency.

Formatted:
The ISO code HUF is used when referring to the Hungarian currency, forint.
Added that as a Currency Unit.


spencermountain:
for @Formatted , this got merge-conflicted with another PR. Do you mind if I just put them in myself?

spencermountain:
that should do it.
cheers!

PR #153, opened 2016-05-20
Update example to show how to get a text response.

blainesch:
It's not obvious there is even a `text` method, I'm guessing because this is an instance of a class.

~~~
> Object.keys(nlp.sentence('I am a sloth').to_past())
[ 'str', 'terms', 'contractions' ]
~~~

If we kept it as `text` it would also have an array syntax:

~~~
> nlp.text('She sells seashells').to_past()[0].text()
'She sold seashells'
~~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **nlp_compromise** does NLP in the browser.
 ```javascript
-nlp.text('She sells seashells').to_past()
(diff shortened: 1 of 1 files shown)


spencermountain:
thanks blaine, this is a good point.
cheers!

PR #161, opened 2016-05-23
Add ISO currency codes and names, fix units

ilyankou:
All units in noun/value/units.js changed to lowercase, as parse() splits the sentence into words and applies toLowerCase(), so "if (units[s])" check made no sense for entries like "EUR".


spencermountain:
woop! thanks!

PR #178, opened 2016-06-01
Fix number for values like [12 6-gram pieces]

### LINKED ISSUE #172: Issue with parsing "6 7-ounce containers"
`nlp_compromise.value("6 7-ounce containers").number = null`
It should be `6`.

ilyankou:
Fix #172


spencermountain:
YOU'RE THE BEST

Issue #127, opened 2016-05-15
Place can also be airport codes (eg JFK)


jashsayani:
The `place` API is interesting, in that the library knowns about places. Not sure if places belongs here but if we have cities, then airport codes can also be very useful.

Example: Find flights from JFK to SFO.

An alternate would be to have "data-plugins" that define sets for categories like places. So an airport plugin can be used that would recognize airport codes as places.

spencermountain:
ah! very neat idea. I like this a lot.
happy to support this one in master.
let me get through some backlogs..
thanks

ilyankou:
Waiting to be merged. The only issue is, since all tokens are returned via filter(), they contain punctuation marks. That is, "Traveling to LAX?".places() will return ["LAX?"].

spencermountain:
yeah, perhaps we need some kind of granular `term.strip_punctuation()`, or something. Very nice ilya

PR #179, opened 2016-06-01
Add airport codes to places

ilyankou:
Implement #127


ilyankou:
I moved "lax" from adjectives to places (airports). If you google "lax", all links on the first page are about the airport.

spencermountain:
nice! this looks great Ilya.

Issue #180, opened 2016-06-01
Noun.conjugate() error with the example from the API documentation


struesdell:
From [the API documentation:](https://github.com/nlp-compromise/nlp_compromise/blob/master/docs/api.md#noun-methods)

```javascript
nlp.noun('mayors of toronto').conjugate()
//{ plural: 'mayors of toronto', singular: 'mayor of toronto' }
```

My results:

```javascript
// NodeJS v5.7.1 REPL
// Babel v6.5.2 w/presets: "es2015", "stage-0"
// nlp_compromise v4.10.5

const nlp = require('nlp_compromise');
nlp.noun('mayors of toronto').conjugate();

TypeError: nlp.noun(...).conjugate is not a function
```

ilyankou:
Conjugation is a term applicable exclusively to verbs. From Wikipedia: 

In linguistics, conjugation is the creation of derived forms of a verb from its principal parts by inflection (alteration of form according to rules of grammar). Conjugation may be affected by person, number, gender, tense, aspect, mood, voice, or other grammatical categories. 
...
The term conjugation is applied only to the inflection of verbs, and not of other parts of speech (inflection of nouns and adjectives is known as declension).

We should add declension() to Nouns doing the same thing as conjugate() in the API documentation.

spencermountain:
hey thanks Shawn, yeah just some old documentation.
that was an attempt to wrap `noun.singularize()` and `noun.pluralize()` into one method, but I agree it's not the right label for it. Right now you can just call each one individually.
@ilyankou how would you feel about a more generic `term.forms()` method? (or something similar)
That way, no matter what the class, you could blast-out all the different transformations we can think of.

2426021684:
Remove noun.conjugate from typings?

PR #184, opened 2016-06-02
Add term.forms()

ilyankou:
#180

### DIFF
--- a/src/term/term.js
+++ b/src/term/term.js
@@ -136,6 +136,21 @@ class Term {
     this.normal = str;
     return this.normal;
   }
+
+  forms() {
+    if (this.pos['Noun']) {
+      return { 'singular': this.singularize(),
+               'plural': this.pluralize()
+             };
+    } else if (this.pos['Verb'] || this.pos['Adjective']) {
+      return this.conjugate();
+    } else if (this.pos['Adverb']) {
+      return {
+        'adjective': this.to_adjective()
+      };
+    }
+  }
+
 }
 
 Term.fn = Term.prototype;
--- a/test/unit_tests/nouns/nouns.js
+++ b/test/unit_tests/nouns/nouns.js
@@ -207,4 +207,21 @@ describe('nouns', function() {
     done();
   });
 
+
+  //Noun forms()
+  it('noun.forms()', function(done) {
+    var tests = [
+                  ['cat', 'cats'],
+                  ['dog', 'dogs'],
+                ];
+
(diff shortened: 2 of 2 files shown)


spencermountain:
hey, wanna put this on the other classes too? LIke, `term.forms()` does: 
```javascript
{text:this.text, normal:this.normal}
```
and then `value.forms()` maybe returns something like,
```javascript
{text:this.text, normal:this.normal, ordinal:this.to_ordinal() ....}
```
etc? I dunno, up to you if you like it.

spencermountain:
i tried overloading `.root()` for the first time, and it's kinda cool. That way people don't need to type-check each term before firing their method. Try it out, if you want. I'm happy to just merge this, if you're not hot on the idea

ilyankou:
I don't quite follow: you can apply .forms() for any term, and depending on the tag it will either conjugate (verbs, adjectives) or return singular/plural for nouns – no type-checking from users. I thought of it exclusively within issue #180. But you would like this operator to basically return everything there is for a term/value?

spencermountain:
yeah, i was thinking we could have .forms() as a method for every class of Term, like: 
* `Noun.forms()` - returns singular/plural forms
* `Verb.forms()` - conjugate, etc
* `Person.forms()` - pronoun, firstname, lastname
to sorta universalize it, in the api - like 'blow-out all derivations' or whatever. That'd be more lucid than fishing around for a `.to_bigLinguisticTerm()` method, you know? Like, `t.show_me_everything()`

spencermountain:
or how about `term.all_forms()` ?

ilyankou:
I created one overloaded method in Term which is inherited by all classes like Noun, Verb, etc (who inherit Term) – I thought it would be easier to manage. Noun.forms() and Verb.forms() hence works exactly as you described above - the "if" statement makes sure the right method (conjugate(), pluralize(), etc.) is called.

ilyankou:
The thing is, for Verb and Adjective forms() basically duplicates conjugate(). But I can definitely implement this method separately in every class with appropriate fields we want to display – although again, my intention was to fix an issue, not to create something new.

spencermountain:
ah, yeah i see. I can merge this now if you want. 
I may rename it and kick it a bit further, to `Person`, `Date` ... and maybe `Sentence`??
unless you wanted to give it a go

ilyankou:
Go ahead and make appropriate changes – and merge when you feel it's ready!

PR #190, opened 2016-06-03
Switch Verb Infinitive negation from `not` to `don't`

brycebaril:
e.g.

walk the dog -> don't walk the dog
call the cops -> don't call the cops

(vs. previously)

walk the dog -> not walk the dog
call the cops -> not call the cops


spencermountain:
:metal: :metal: :metal: :metal:

PR #192, opened 2016-06-04
Remove debug console.log statements that I forgot to remove

2426021684:
Remove debug console.log statements that I forgot to remove


spencermountain:
in the future, don't worry about running a build for a pr. we can just crank away in ./src, then fire off a build each release.
cheers

PR #199, opened 2016-06-07
Add grunt-cli dev dependency

nathanmarks:
This allows use of the `npm` run scripts (or even just grunt directly) without having grunt-cli installed globally. 

This is also what the [grunt repo recommends](https://github.com/gruntjs/grunt-cli):

> **Note:** The job of the grunt command is to load and run the version of Grunt you have installed locally to your project, irrespective of its version. Starting with Grunt v0.4, you should never install Grunt itself globally. For more information about why, [please read this](http://nodejs.org/en/blog/npm/npm-1-0-global-vs-local-installation).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "natural language processing in the browser",
   "version": "6.1.2",
   "main": "./builds/nlp_compromise.js",
-  "browser":"./src/index.js",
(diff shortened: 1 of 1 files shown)


spencermountain:
thanks!

PR #201, opened 2016-06-07
Keep quote consistency in documentation

MrChristofferson:
All single quotes for function params/implied string outputs and backticks for escaping :)


spencermountain:
thanks Chris

Issue #198, opened 2016-06-06
Noun/Verb being tagged as Adjective


nathanmarks:
I noticed that certain nouns / present participle verbs get tagged as an adjective, for example:

> I work in advertising.

Gives me:

```js
Person {
  whitespace: { preceding: '', trailing: ' ' },
  text: 'I',
  normal: 'i',
  expansion: null,
  reason: 'is_person',
  pos: { Noun: true, Plural: false, Person: true, Pronoun: true },
  tag: 'Person',
  honourific: null,
  firstName: null,
  middleName: null,
  lastName: null }
Verb {
  whitespace: { preceding: '', trailing: ' ' },
  text: 'work',
  normal: 'work',
  expansion: null,
  reason: 'lexicon_pass',
  pos: { Verb: true, Infinitive: true },
  tag: 'Infinitive' }
Term {
  whitespace: { preceding: '', trailing: ' ' },
  text: 'in',
  normal: 'in',
  expansion: null,
  reason: 'lexicon_pass',
  pos: { Preposition: true },
  tag: 'Preposition' }
Adjective {
  whitespace: { preceding: '', trailing: '' },
  text: 'advertising.',
  normal: 'advertising',
  expansion: null,
  reason: 'rules_pass_63',
  pos: { Adjective: true },
  tag: 'Adjective' }
```

nathanmarks:
Here's a failing test I'm toying with: https://github.com/nathanmarks/nlp_compromise/commit/f137548153ba3d238f16d598e4ba62fa77e51d2c#diff-95ea7ae895a643b8ed30173e1ea8624fR37

spencermountain:
fellow Torontonian!
thanks Nathan, I'll check it out today

nathanmarks:
Hah, fancy that!

It triggers on this regex rule in `pos/passes/rules/word_rules.js`:

```
  ['.[pdltrkvyns]ing$', 'JJ'],
```

nathanmarks:
I should have caught onto you at `canadian bacon` 😉

ilyankou:
Resolved, waiting to be merged. (#204)

Issue #194, opened 2016-06-04
"4 cats" and "4 red cats" are parsed very different


2426021684:
"4 cats" => Value("4 cats")
"4 red cats" => Value("4"), Adjective("red"), Noun("cats")

Why are they parsed so differently?

spencermountain:
yeah, this is a tough case.
i think in the future, we'll want `"4 cats" => Value(4), Noun('cats')`
then our value parsing method jumps around the sentence, interpreting `.of_what()` etc.

2426021684:
I agree

2426021684:
Or maybe have hierarchical terms. Have value contain `Value(Number(4), Noun(Adjective("red"), Noun("cats")))`

ilyankou:
Resolved, waiting to be merged. (#204)

Issue #193, opened 2016-06-04
Roses incorrectly marked as verb


2426021684:
In `nlp.text("4 red roses").terms()`, roses is a verb. Roses should be a noun in this case because it comes after an adjective.

2426021684:
```
> nlp.text("five kittens").terms()
[ Value {
    whitespace: { preceding: '', trailing: ' ' },
    text: 'five',
    normal: 'five',
    expansion: null,
    reason: 'is_value',
    pos: { Noun: true, Value: true },
    tag: 'Value',
    number: 5,
    unit: '',
    unit_name: null,
    measurement: null,
    of_what: '' },
  Verb {
    whitespace: { preceding: '', trailing: '' },
    text: 'kittens',
    normal: 'kittens',
    expansion: null,
    reason: 'rules_pass_37',
    pos: { Verb: true },
    tag: 'Verb' } ]
```
Kittens should be marked as a noun because it follows a number.

ilyankou:
Resolved, waiting to be merged. (#204)

PR #204, opened 2016-06-11
Fix parsing + some code refactoring

ilyankou:


ilyankou:
I removed a rule from `lumper.js` that turns `Value + Noun` into a `Value`. That is, previously `5 books` was marked as a Value, now it is `['Value', 'Noun']`. I also added two rules to `grammar_rules.js` that turn `Value + Verb` into `Value + Noun` ('5 kittens'), and `Adjective + Verb` into `Adjective + Noun` ('red roses').

Fixes #205, #198, #194, #193

test/unit_tests/sentence/tagger.js
```diff
@@ -65,3 +66,6 @@ describe('custom lexicon', function() {
   });
 
 });
+
+console.log(nlp.text('3 buses').terms());
```
spencermountain: heads up ☝️

spencermountain:
yeah, thanks ilya, if it made it easier, you could also coerce '5 trains' into a `Value`, although I like what you've done.
just comment those console.logs and we're good-to-go

ilyankou:
Fixed!

spencermountain:
hey, just pushing a build and wanted to check something with you
[this `w` variable is redeclared](https://github.com/nlp-compromise/nlp_compromise/blob/master/src/term/noun/value/value.js#L251), 
can i just rename it to something else?
(caught it in the linter)

ilyankou:
Sure go ahead!

PR #209, opened 2016-06-14
Improve name parser to support double first/last-names, proper surname case

### LINKED ISSUE #206: Person name fields inappropriately lower cased
Version: 6.1.2
```
var nlp = require('nlp_compromise');

console.log(nlp.text('Ronald McDonald').people());
```

```

[ Person {
    whitespace: { preceding: '', trailing: '' },
    text: 'Ronald McDonald',
    normal: 'ronald mcdonald',
    expansion: null,
    reason: 'is_person',
    pos: { Noun: true, Person: true },
    tag: 'Person',
    honourific: null,
    firstName: 'ronald',
    middleName: '',
    lastName: 'mcdonald' } ]
```

I have no way to get the last name with correct capitalization.

ilyankou:
Resolves #206


spencermountain:
nice one! love it.
can `lastname_case` and `title_case` stay off the `Person` class? Otherwise they're exported publically. Just throw them at the top of the file or something, if you can.

lastly, would it make sense to coerce to lowercase in Person.root()? I had imagined .root() would make very-machine-readable text, but it's kindof half-baked right now.

Love the double-name thing. Never thought of that. That's brilliant.

ilyankou:
Done!

For the `root()`, I simply made it return `toLowerCase()` from `firstName`, `middleName`, and `lastName`, because they had been properly normalized before being processed.

PR #212, opened 2016-06-16
Use scientific notation for clarity.

Kiran-Rao:


spencermountain:
... and filesize! thanks Kiran.

PR #223, opened 2016-06-19
Add additional demo

brandonJY:
Demo to show the current API capability (not all API is included). 
And it can be used for testing purpose or as a handy tool for other purposes such as grammar checker.


spencermountain:
thanks!

PR #224, opened 2016-06-20
Add words to lexicon with a JSON-like object

ilyankou:
```javascript
let lex = nlp.lexicon({'apple': 'Person',
                       'jackie': 'Adverb',
                       'amazing': 'Noun'
                       });
```


spencermountain:
ah, yeah, this is a way better api for that. nice one!

PR #227, opened 2016-06-21
Add all_forms() to adj, adv, noun, and verb

### LINKED ISSUE #222: adverb.conjugate() is not implemented
I saw nlp.adverb.conjugate which is mentioned in API doc https://github.com/nlp-compromise/nlp_compromise/wiki/API#adverb-methods
has not been implemented. 

Currently only nlp.adverb.to_adjective has been implemented. Is there a plan to wrap it with conjugate function in the future? Or should we just update the API docs?

ilyankou:
Fix #222 via all_forms(), not conjugate().

### DIFF
--- a/src/term/adjective/adjective.js
+++ b/src/term/adjective/adjective.js
@@ -36,10 +36,16 @@ class Adjective extends Term {
       noun: adj_to_noun(this.normal)
     };
   }
+  all_forms() {
+    var forms = this.conjugate();
+    forms['normal'] = this.normal;
+    return forms;
+  }
 
 }
 Adjective.fn = Adjective.prototype;
-// let t = new Adjective("quick")
-// console.log(t.conjugate())
+
+//let t = new Adjective("quick")
+//console.log(t.all_forms());
 
 module.exports = Adjective;
--- a/src/term/adverb/adverb.js
+++ b/src/term/adverb/adverb.js
@@ -11,9 +11,16 @@ class Adverb extends Term {
   to_adjective() {
     return to_adjective(this.normal);
   }
+  all_forms() {
+    return {
+      adjective: this.to_adjective(),
+      normal: this.normal
+    }
+  }
 }
 Adverb.fn = Adverb.prototype;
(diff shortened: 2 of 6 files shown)


ilyankou:
@spencermountain There is definitely a way to combine two JSON-like objects into one like that:
```javascript
function merge_options(obj1,obj2){
    var obj3 = {};
    for (var attrname in obj1) { obj3[attrname] = obj1[attrname]; }
    for (var attrname in obj2) { obj3[attrname] = obj2[attrname]; }
    return obj3;
}
```
Credit: http://stackoverflow.com/questions/171251/how-can-i-merge-properties-of-two-javascript-objects-dynamically

So for the Person, you want it to combine the info of the Term (like `normal`), then Noun (plural, singular), and then Person?

spencermountain:
yeah, i mean like, when two methods are overloaded, like `all_forms()`, they all call their parent somehow.
It [may](http://stackoverflow.com/questions/32615034/call-parent-function-which-is-being-overridden-by-child-during-constructor-chain) be possible? 
```
Person.all_forms(){
  return merge_options(super.all_forms(), {new_form : 'blah' })
}
```
:shrug:

spencermountain:
what i'm thinking is, if we added a result in `Term.all_forms()`, we wouldn't need to re-add it in all the other classes - just saving us some trouble maintaining the class hierarchy stuff.

ilyankou:
@spencermountain Yes, I understand. Utilizing OOP at its full! Will look into it

kittykatattack:
Hey guys, we to update the docs! :smile: 
`conjugate` is still all over the place, including, incorrectly, for `adverb` - which is how I stumbled upon this issue.

spencermountain:
sorry about this. Yeah, [the new version](https://github.com/nlp-compromise/nlp_compromise/tree/v7) is a dramatic change, and I will delete/rewrite the docs+wiki.
I hate not fixing things, but the new version is wicked, and well on its way.

kittykatattack:
@spencermountain Thank you, Sir! 👍

PR #228, opened 2016-06-21
Add support for date abbreviations jun, jul and sep

kahwee:
I notice that jun, jul and sep aren't there. I added them in.


spencermountain:
thanks!

PR #234, opened 2016-06-24
Update build

nloveladyallen:


spencermountain:
hey noah, is this just for the Basque change? 
i haven't got the build flow down yet, but it makes sense for me to merge un-buillt prs, then build on a release?

nloveladyallen:
Yes, just for Basque. 

I generally think the build should be updated along with the source (thus this PR), but it's not my project. ¯\\\_(ツ)_/¯

PR #237, opened 2016-06-27
You use the symbol `฿` but not the colloquial name `baht`

soyjavi:


soyjavi:
Thanks @spencermountain I'm living in Thailand so I'm quite sensible with this country :)

PR #245, opened 2016-07-19
Correctly detect acronyms with 2 or more letters

stevebakh:
This change updates the match to correctly detect acronyms that are made up of only 2 letters, such as UN, EU, AA, etc.


spencermountain:
thanks!

PR #273, opened 2016-10-02
Added Street Suffixes and Cleaned Up Duplicate or Mis-Spellings

davidbuhler:
1. - Added suffixes from USPS Common Road Suffix Names (see http://pe.usps.gov/text/pub28/28apc_002.htm)
3. - Provided line-sorting on some lists, to remove duplicates values (removed duplicate first names).
4. - Addressed spelling errors with Street Prefixes and added missing Street Prefixes.


spencermountain:
ah, very clever. Thanks david!

davidbuhler:
NP.

I separated out the common mis-spellings into their own collection, as well as the abbreviations (perhaps they can be part of a plugin). 

The abbreviations  list is likely to cause false positives. It's also possible some of the USPS suffixes might create false positives. I tried to look at how Gate and ANNNIE resolve the issue when performing entity extraction for Place/Location, which (I think) they resolve by looking at patterns for tokens in the near proximity (if it's a street suffix, is there a unit number nearby?). 

If the suffixes do return false positives, they might need to be whittled down or pattern matched based on proximity.

spencermountain:
hey, `common_street_designators` and `street_designator_abbreviation` didn't seem to be being used anywhere. i commented-out them out for now, if you wanna integrate them in is_place.js, giver!

PR #275, opened 2016-10-04
Adds additional organizations from OpenData

davidbuhler:
Added organized from http://www.opendata500.com/us/list/

Some entries might need suffixes removed (-software, -group) to show up in conversational text.


spencermountain:
hey david, great idea beefing up the organizations - is there any way to make this smaller?
The client-side use-case forces things to be really tiny. 

Looking at them, some are obscure, like 5pSolutions, others are already terms in the lexicon, like 'factual'. And some are noisy terms, like 'farmers' - any other datasets you can find?
...like, companies by brand popularity, or something?
cheers!

davidbuhler:
Hi Spencer,

What if the organizations were broken down into:

1) top 100 companies
2) top 20 government agencies
3) top 20 non-profit organizations

I can structure like above if you prefer, or I'm open to other approaches.

In my use-case, I need as many US organizations as I can get to extract Place entities, and I need as many first names last names as possible to extract Person entities.

Is there a way to use the plug-in architecture to add specific Nouns to the Noun list for Organizations? If so, I can move some datasets into plug-ins.

spencermountain:
hey, yep of course. You can set the lexicon [to anything you'd like](https://github.com/nlp-compromise/nlp_compromise/wiki/Lexicon#adding-words)

ya definetly, the current list now really sucks. I think i just spit a few out by hand. Lots of room to improve this.

davidbuhler:
Hi Spencer,

How about if I do:

Top 100 companies
Top 50 non-profits
Top 50 government agencies

I can do a cross reference against generic terms.

- How does disambiguation work within the framework? Is there a disambiguation utility? CoreNLP uses http://nlp.stanford.edu/fsnlp/wsd/, I believe, which is probably a framework unto itself :)
- Is there a way for me to write a plugin that extends organizations?

davidbuhler:
Sorry about the double-post :S

PR #289, opened 2016-10-25
$ signs are filtered out on numbers, and grand is interpreted as 1000

Abazhenov:
$ signs are now filtered out similar to hashtags and and at mentions. Grand is now interpreted as '1000'


spencermountain:
beauty! thanks!

PR #290, opened 2016-10-25
term.toTextNumber() now works with positive numbers

Abazhenov:
We can now call toTextNumber directly off of a term.  All test related to toTextNumber except for the commented out -5 are passing.


spencermountain:
thanks! yeehaw.

PR #294, opened 2016-11-10
Fix result/render/asArray method

anastasiia-zolochevska:


spencermountain:
you're great! thanks Anastasia

Issue #302, opened 2016-12-19
The word invest/ invested is not recognized as a verb


geekodour:
Adding it to the lexicon solves the issue but, this is something that should not have happened because other verbs are almost working fine. Any idea why?

spencermountain:
hi Hrishikesh, thanks, i'll take a look.

PR #303, opened 2016-12-19
fix(verb) Fix for issue #302: The word invest/ invested is not recognized as a verb

TSavo:
Include 'invest' as a verb in the data. Add additional unit tests to show that invest is treated as a verb.


spencermountain:
boom!
thanks Travis!

PR #307, opened 2017-01-10
Add compatability for environments where self is not defined

AndrewGHC:
This is a small change that enables the library to be used in environments that have not defined 'self'. Currently, this library uses 'self' rather than 'window' to ensure compatibility with web workers (see PR https://github.com/nlp-compromise/nlp_compromise/issues/72). This PR keeps this functionality but ensures that everything works in non-browser environments such as React Native where self is undefined.


spencermountain:
Very clever! Thanks! 
Is it working on react native? Would love to hear what your experience has been.
I'll also add this change to the new upcoming v7 branch. 
Cheers

AndrewGHC:
It is indeed, I'm using it in a very simple fashion just to play around with react native and it works perfectly with this patch. Great work on this library, it's good fun to use!

PR #317, opened 2017-02-10
Added more countries, we can now detect at least 231 countries.

wallali:
Used the list at http://www.geonames.org/countries/ to add several more countries to the list.


spencermountain:
nice! thank you!

Issue #261, opened 2016-08-31
Cannot build in Windows


leoseccia:
Hello, 

I am very new to Node so I might be doing something wrong...

I'm on Win 7 x64 and I ran the following successfully:
`nvm use 4`
`npm install`

However when I attempt to run:
`npm run build` or `npm run build_windows`

I get the following error:

```
c:\Users\seccial\node source\nlp_compromise-master>npm run build

> nlp_compromise@6.5.0 build c:\Users\seccial\node source\nlp_compromise-master
> grunt build

Running "run:test" (run) task
'.' is not recognized as an internal or external command,
operable program or batch file.
Warning: non-zero exit code 255 Use --force to continue.

Aborted due to warnings.

npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\
node_modules\\npm\\bin\\npm-cli.js" "run" "build"
npm ERR! node v4.0.0
npm ERR! npm  v2.14.2
npm ERR! code ELIFECYCLE
npm ERR! nlp_compromise@6.5.0 build: `grunt build`
npm ERR! Exit status 3
npm ERR!
npm ERR! Failed at the nlp_compromise@6.5.0 build script 'grunt build'.
npm ERR! This is most likely a problem with the nlp_compromise package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     grunt build
npm ERR! You can get their info via:
npm ERR!     npm owner ls nlp_compromise
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     c:\Users\seccial\node source\nlp_compromise-master\npm-debug.log

```

and

```

c:\Users\seccial\node source\nlp_compromise-master>npm run build_windows

> nlp_compromise@6.5.0 build_windows c:\Users\seccial\node source\nlp_compromise
-master
> grunt build_windows

Warning: Task "build_windows" not found. Use --force to continue.

Aborted due to warnings.

npm ERR! Windows_NT 6.1.7601
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\
node_modules\\npm\\bin\\npm-cli.js" "run" "build_windows"
npm ERR! node v4.0.0
npm ERR! npm  v2.14.2
npm ERR! code ELIFECYCLE
[...]

PR #320, opened 2017-02-14
Fix build on windows #261

wallali:
Fixes issue #261 
Please check before merging that this still works on *nix


spencermountain:
yes!! thank you Ali.

PR #332, opened 2017-03-15
Master

leoseccia:
Small change to allow multiple tags to be passed to tag methods as an array


leoseccia:
Hello,

This is a very simple sample of the object I pass to this step:
```
  var testCustomRules = {
        singleTermRegularExpressions: //like in emoji step - regX match // this is the first pass
        [{
            description: '4-digit-year', //this is just ref...
            //normalize: { from: 'a', to: 'b' }, // this is a text replace
            regX: /^(\(|\[)?[1-2]\d{3}(\)|\])?$/, // /\b(\(|\[)?[1-2]\d{3}(\)|\])?\b[^-]/,
            tags: ['Legal', 'LegalFourDigitYear']
        },
        {
            description: 'year-range', //this is just ref...
            //normalize: { from: 'a', to: 'b' }, // this is a text replace
            regX: /^(\(|\[)?[1-2]\d{3}-(?:\d{2}){1,2}(\)|\])?$/,
            tags: ['Legal', 'LegalYearRange']  
        }
        ],
        lumpExpressions: //like in dates (ie. tags # instead of regx match) //this is a second pass
        [{
            description: 'lumpRule1',
            match: '#Person #Legal',
            tags: 'PLegal1' 
        },
        {
            description: 'lumpRule2',
            match: '#Verb #Legal',
            tags: ['PLegal3', 'PLegal4']
        }
        ]
    };
```

spencermountain:
hey Leo, this is neat. Thanks for doing this.
let me spend some time pondering on this.
I like the idea of passing-in more controls for tagging - what's the advantage for doing this instead of tagging 'in post'? Users could always run this:
```js
myTaggedText.match('#Verb #Legal').tag('PLegal3').tag('PLegal4') //....
```

hey, come to think of it, since you're extending the tagset here too, maybe that oughta come into this somehow. I tried using that 4rth parameter a few weeks ago to somehow to extend the [tag-heirarchy](https://github.com/nlp-compromise/compromise/blob/master/src/tags/tree.js) so that `PLegal4` defines `PLegal3` or whatever too.
any thoughts on that?

spencermountain:
maybe the custom regex is a thing in itself too. And the lumper. That's pretty powerful

spencermountain:
if we had user-defined lumping of two terms, i'd prefer it lived as an api - something like:
```js
nlp('dr John was cool').match('dr #Person').lump().terms(0).out() //'dr John'
```

leoseccia:
Hey Spencer,

Yes, all this stuff is not finalised or by any means production ready, I just wanted to show you how I am poking and changing compromise to just find out more about it.

At this particular point, my main need was to use regEx to tag terms... 

It would be great to have a plugin architecture :p (only joking)

I love the idea of a lumping method in the api. 

Tag hierarchy is a great idea but i really didn't spend any time looking at it. Will do soon.

PR #335, opened 2017-03-17
Tag using tag arrays and lumping methods + extending tag hierarchy by passing tagset to nlp constructor

leoseccia:
Allow tags array to be passed to tag methods
Added some lumping methods


leoseccia:
...this is how i use it...

`ts.match('#Noun #Verb').lumpIntoParent(['myTag1', 'myTag2'], 'reason i am tagging it...');`

leoseccia:
Added a way to extend tagset hierarchy... ie.  
nlp("My text to parse", myLexicon, {LegalParent :{ Legal: true}});

spencermountain:
nice! I couldn't figure that out.
so we're not going with the `customRules` thing anymore? The third param is now an object that extends the curent PartOfSpeech tags? 
:fireworks: 
wanna add a few tests, then I'd be happy to merge this one in

spencermountain:
really nice work Leo.
CC @owendall

spencermountain:
ah rats, there's a couple tests failing. I tried looking at them but couldn't figure it out.
Gotta revert this. :cry:

spencermountain:
weird that travis-cli didn't catch them.
also, heads-up `builds/` is in .gitignore

leoseccia:
hey Spencer,

Sorry i didn't notice these... I think the issue is with the change in src/tagger/steps/02-lexicon_step.js and the fact that in our current tagset we have incompatibilities between tags eg. City cannot be a Person. 

I'd like to fix this by allowing a term to be both terms ie. Paris is both a City and a Person. When you say Paris is nice - this could refer to either the city or the person

The alternative is to revert my change in this step and go back to having the custom lexicon overriding the built in one. 

What do you think?

Leo

PR #341, opened 2017-03-22
REDO: Tag using tag arrays and lumping methods + extending tag hierarchy by passing tagset to nlp constructor

leoseccia:
Allow tags array to be passed to tag methods
Added some lumping methods - this is how i use it: ts.match('#Noun #Verb').lumpIntoParent(['myTag1', 'myTag2'], 'reason i am tagging it...');
Added a way to extend tagset hierarchy... ie.
nlp("My text to parse", myLexicon, {LegalParent :{ Legal: true}});
Relaxed tag conflicts - some names can be both Male and Female and these are not conflicting - 'Person', 'Organization', 'Value', 'Place', 'Actor', 'Demonym', 'Pronoun'
-All unit tests pass now


leoseccia:
closed because Spencer does not want to relax Conflicts and would rather have each change in a separate pull request

PR #345, opened 2017-03-23
Lump methods

leoseccia:
Added some extra lump methods to allow lumping inline ie. ts.match('#Noun #Verb').lumpIntoParent(['myTag1', 'myTag2'], 'reason i am tagging it...')

This is useful for the API end users who want to lump their match results


spencermountain:
hey, i've made some changes to how this behaves that you may wanna look over,
I combined ('lumped'?) `.lump()` and `.lumpIntoParent()` into one method, so that it doesn't over-reproduce the functionality of `.tag()`. Used in combination, it looks like this:
```js
var r = nlp('we live in Toronto Canada and it is cold');
r.match('#Place+').lump().tag('MyTag', 'myreason')
//['we', 'live', 'in', 'Toronto Canada'...]
```
lemme know if that works for you?

spencermountain:
if this breaks the other pr, i'll fix it

spencermountain:
so another thing that's cool about this is that the tagger itself can use it, when it does its lumping. 
i'm moving that logic into your directory now, gimme 2 mins.

PR #348, opened 2017-03-24
Tag hierarchy + tagging improvements

leoseccia:


leoseccia:
This contains a few changes to do with tagging...

First of all, it adds the ability to pass a tag hierarchy tree to nlp to extend the built in one. NB. not sure how this works with existing tags (as in already existing in the built in tagset) - i used it for new tags only

It then also allow terms to be tagged by multiple entries in the lexicons/tries. This caused some issues with the unit tests, so there is also some code to remove tag conflicts - first tags always take priority, therefore custom lexicon first, then built in, then tries.

spencermountain:
hey leo, been working really hard on cleaning up this part of the stack. tagging is definetly too slow right now for what it does - been also thinking that extending tags could be done with a method call somehow, instead of all the extra-parameter-stuff. - like nlp.tagSet(tags) or something maybe?
what do you think?
anyways, there's already conflicts with this, and I'm just about to make some more.
whatever we chose, this must be tested really hard.

leoseccia:
Hi Spencer,

Are you sure that this increase execution times? For the existing test cases, it should be minimal since a tagset is not passed and the functionality is (should be) bypassed. I guess there are some changes in the lexicon step which might affect performance...

I agree that the pos tag process is rather slow. Maybe we should consider merging the steps iterations (going through terms) across 2 or 3 passes (rather than having one iteration per step)... Not sure how much it would help but worth a try. Did you do much perf profiling? What did you find? Where are the bottlenecks?

I haven't looked at the matching methods in any detail yet... How do they perform?

For tagging in particular, I still think that we are much better off taking an inclusive approach (eg. tag the name Mel as MaleName, FemaleName and PersonName) and minimise logic around it - I would confine the conflict resolution stuff strictly to enforce grammar rules.

Anyhow, I am going to do some perf. profiling too at some point... I really want to but I am just quite busy with paid work right now. I'll let get in touch, if I get any more ideas or improvements.

From a usage point of view, I agree with you and I think all the extra parameters in the main method nlp() are a bit much... I would remove skipTagging and have a separate tokenize method, I would then merge the lexicon with the tagset into a json serializable object (and possibly include other user data in it in the future - things like custom rules or tag conflicts) and pass that as a single optional parameter.

Speak to you soon. Please don't feel obliged to take any of this stuff into the main branch and thanks for taking it into consideration.

Leo

PR #403, opened 2017-08-22
QuestionWord and EndPunctuation fix

sebilasse:
_QuestionWord_
 The words `how` and `which` are fairly often `QuestionWord`, also when
used for indirect questions.
As described in https://github.com/nlp-compromise/compromise/issues/399

_EndPunctuation_
EndPunctuation fixes for spaces, consecutive pointings and combinations
as described in https://github.com/nlp-compromise/compromise/issues/397

- tested fix, all available tests passed


spencermountain:
beauty! 
thanks.

PR #421, opened 2017-10-31
adding Irish names

kanejoe:
adding some specific Irish names, but many international.  Would find v useful for my needs.  Many thanks


spencermountain:
thanks Joe!

hey, you may be interested in using the (very new) [compromise-plugin](https://github.com/nlp-compromise/compromise-plugin) feature.

you can see one of the examples i had in mind when making it was a (yet uncreated) scottish-slang plugin, which would look something like this:
image

try it out, if you'd like,

cheers

kanejoe:
I certainly will check that out.  Well done on the project, really good.  I'll have some more names to add shortly too.

Issue #254, opened 2016-07-29
typings file


adamjuhasz:
The published package is missing the type file `nlp.d.ts`

``` bash
Adams-MacBook-Pro-5:nlp_compromise adamjuhasz$ ls -l
total 32
-rw-r--r--   1 adamjuhasz  staff  1080 Jan 27  2016 LICENSE.txt
-rw-r--r--   1 adamjuhasz  staff  4029 Jul  6 07:33 README.md
drwxr-xr-x   4 adamjuhasz  staff   136 Jul 27 08:42 builds
-rw-r--r--   1 adamjuhasz  staff  2626 Jun 20 17:38 changelog.md
-rw-r--r--   1 adamjuhasz  staff  2793 Jul 27 08:42 package.json
drwxr-xr-x  10 adamjuhasz  staff   340 Jul 27 08:42 src
```

spencermountain:
yikes! thanks Adam! it's [right there](https://github.com/nlp-compromise/nlp_compromise/blob/master/nlp.d.ts), but maybe it's not being shipped with npm, because of the package.json settings.
Honestly, I've not figured out how to test this typefile format, and get lost whenever I research it.
What are you using it for?

adamjuhasz:
I would have made a PR, but not sure how to fix this, it's in the repo and in package.json:typings so it must be some configuration setting for NPM... I thought someone here might know a quick answer, but I'll do a bit of googling.

I'm using nlp_compromise with our chatbot framework (https://github.com/fyndme/bot-framework) currently I just grabbed the d.ts file and threw it in manually.

adamjuhasz:
@spencermountain, does it need to be added to the 'files' section of package.json?
#255 should do it

spencermountain:
thanks man, do you have any feelings about jsdoc/flow annotions?
i'm working on a major api change and may add type annotations, and some auto-documentation, but haven't landed on which style to jump on yet.

spencermountain:
the separate typings file like that is a bit of work, and not easy to like, validate the tests on or whatever.

TheShashank:
wait so is there a typings file?
I did `typings search nlp` and `typings search compromise` but both returned with no results.
I'm loving the browser-based implementation.

spencermountain:
thanks!
yeah, hah. well I'll admit I've looked and still haven't the slightest idea how to do this properly ;)
[...]

PR #433, opened 2017-12-01
Make it easier for Typescript to load types

WorldMaker:
Related to #254; this quick change lets Typescript load the current "tentative" types file during node module resolution.


spencermountain:
thanks!

PR #438, opened 2017-12-16
Add responses via plugin

playground:


spencermountain:
hey there, thanks for the work. When you're doing a pr, please describe what the change is doing and why you're doing it, so I can better evaluate it.
there doesn't seem to be a addResponses.js file, so the build is breaking.
there are also several files that have been changed that shouldn't need to be.
thanks

playground:
Hi @spencermountain, sorry I haven't touched this for over a month and I was working off of an outdated branch and also the wrong repo.  Here it's again.

I added addResponses.js to provide the ability to add additional supports of command & control via plugin.  For now, I have "navigation" & "weather" as an example to determine the intent of a sentence/question if it is about wanting to know about the weather or navigation.  Based on the resulting intent one can provide logic to provide corresponding response.  Please let me know if there is a better way to perform this sort of operations.  My goal is to extend this further into conversational.  Do you have any examples demonstrating conversational functionalities?

Thanks,

spencermountain:
hey, thanks for this. This is an interesting idea to help classify sentences using the match syntax in a plugin. Unfortunately I don't know if compromise is ready for sentence/statement/clause level tagging or classification. Sorry if I've misunderstood the changes, but adding support for chat-bot level responses is probably outside the scope of the library for now.
have you seen [superscript?](http://superscriptjs.com/) it's using compromise, and adding a layer of call-response logic on-top.

heads-up too, a lot of the match syntax i see in this pr is more-advanced regex than is supported, things like `rain[ing|y]` won't catch those words, as we're not quite that fancy. you can opt-in to a regex, [like this](https://runkit.com/spencermountain/5a394181ec8a44001242ae97), but remember this will be run on every word in your document, so it may just be faster to list-out `(rained|raining)`, etc.
lemme know if i can help,
cheers

playground:
Hi,

Adding a layer on top of Compromise is what I have in mind.  The changes
for add-response is meant to help determine/classify user intent so we can
handle the response in our own logical layer.  rain[ing|y] seems to be
working for me.

spencermountain:
hey thanks for this.
gonna close it, as it merge-conflicts.
happy to work something out
cheers

PR #441, opened 2018-01-03
Handle case where a sentence has no terms

holm:
We ran into this case in production. I think this is the right way to fix it.


spencermountain:
nice! thank you Christian!

Issue #442, opened 2018-01-08
Slow processing of long strings of digits


winstonewert:
The following code

> nlp = require('compromise')
> nlp("101010101010101010101010101010101010101010")

Takes a long time to complete.

This is ultimately due to the regex here:

https://github.com/nlp-compromise/compromise/blob/c9f2a49a137cfe4bb06cb73567144e169d941a04/src/tagger/steps/06-suffix_step.js#L19

/^([0-9]+[,\.]?)+(st|nd|rd|r?th)$/

I believe the regex can be rewritten as:

/^[0-9][0-9,\.]*(st|nd|rd|r?th)$/

To match the same content but avoid exponential backoff.

Additionally, the continue statement here: 

https://github.com/nlp-compromise/compromise/blob/c9f2a49a137cfe4bb06cb73567144e169d941a04/src/tagger/steps/06-suffix_step.js#L76

has no effect because the continue is the last statement in the loop. Given the structure of the function, it is supposed to be a break?

spencermountain:
hey! thanks. 
yeah, the continue thing, i'm sure it's just me being sloppy. 
codacity found [5 other](https://www.codacy.com/app/spencerkelly86/compromise/issues/index?&filters=W3siaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6WyJTZWN1cml0eSJdfV0=) regexes that it said were logically problematic, and I've tried re-writing them to be better-big-O, but couldn't figure it out. wanna take a crack at the others?
thanks a bunch.
cheers

winstonewert:
I'm not really an expert on regexes, but none of the ones pointed out there seem problematic to me. I think they may be false positives for the test.

spencermountain:
think we got all the problematic ones. Good eye, and let me know if you see any other good finds.
thanks!

PR #443, opened 2018-01-08
Fix for #442

winstonewert:
Fixed exponential backoff by rewriting regex.


spencermountain:
you are the best. thank you Winston

Issue #533, opened 2018-10-25
False positive when detecting questions


shamoons:
https://runkit.com/shamoons/false-positive-compromise

```
const nlp = require('compromise')

let res = nlp('Let me run get a price take about 5-10  minutes to process and then I\'ll send you what I got.').sentences().isQuestion().out('array')

console.log(res)
```

This is being detected as a question, when it is evidently not.

spencermountain:
nice find!
Relevant code is [here](https://github.com/spencermountain/compromise/blob/master/src/subset/sentences/isQuestion.js) if anyone wants to jump on this. Otherwise, happy to take a look

shamoons:
I'm on it

shamoons:
@spencermountain - when I `console.log` to debug, nothing prints out when running `yarn test`. Any tips?

spencermountain:
@shamoons oh right, i think that test-reporter disables them, you can see console logs if you do
 `yarn test:spec`
i should fix that...

PR #534, opened 2018-10-25
Fixing isQuestions

shamoons:
https://github.com/spencermountain/compromise/issues/533

Not ready yet. Any thoughts?


shamoons:
cc @spencermountain

spencermountain:
hey shamoon, yeah that's fine. one note that `.has()` doesn't support full-regex, [here](https://github.com/spencermountain/compromise/blob/d5d1a17485c35b3896c2e87c1527d825a4766d76/src/subset/sentences/isQuestion.js#L35)
lemme know when it's ready.

shamoons:
@spencermountain funny you should say that. I was having an impossible time with `.has`, detecting `..` at the end of sentences. Any suggestions on how to better handle it?

spencermountain:
oh, good point. ergh - yeah, this may not be possible right now.
the elipses is stored in the 'whitespace', so they can't be queried with regex, or anything.
we should tag them as `#Elipses`. i'll [make an issue](https://github.com/spencermountain/compromise/issues/535) for this.

shamoons:
Any way for me to match with regular regex? Can `ts` be converted to a string?

shamoons:
cc @spencermountain

spencermountain:
yeah, the regex match syntax is documented here
https://beta.observablehq.com/@spencermountain/compromise-match-syntax

shamoons:
@spencermountain can I run `nlp(ts)` around here? https://github.com/spencermountain/compromise/pull/534/files#diff-1fe477e9e488dda77b528c63b7ab9adbR17

spencermountain:
i'll add support for tagging the Elipses in v12, don't worry about that for now.

if you needed to do a regex, you could do `ts.out('text').match(reg)` re-parsing the whole object should not be necessary.

PR #566, opened 2019-02-10
s/comma-emoji/colon-emoji

oaguy1:
When looking over the different tagger steps, I noticed that colon emojis were improperly tagged as "comma-emojis." This PR seeks to resolve that.


spencermountain:
ha! good find. thank you.
oops!

Issue #568, opened 2019-02-27
Unclear how to convert present tense verb to past participle


moberemk:
So, I'm trying to figure out how to normalize user data to use as part of our interface. One example I'm having particular trouble with is one example, where the user would enter "take a shower" as their text.

Ideally, I want to convert this to a question ("Have you taken a shower?") but I'm not clear on how to convert "take" to its past participle "taken" in this context. Some practical examples:

```
> nlp('take a shower').sentences().prepend('have you').append('today').toPastTense().toQuestion().out
('normal')
'had you take a shower today?'

> nlp('take a shower').sentences().prepend('did you').append('today').toPastTense().toQuestion().out(
'normal')
'did you take a shower today?'
```

The second example, prefixing the sentence with "did you", works as expected, but the first example both fails to convert "take" to the right conjugation as well as incorrectly changes "had" to "have".

I can make this work, but looking at the published API I don't see how to convert "take" to the proper tense manually, and none of the functions I'd expect to do it for me automagically (`toQuestion`, `toPastTense`) are doing it correctly, at least, in the call order that I'm using.

(If this is out-of-scope for the library or if I'm using the API wrong, please let me know!)

spencermountain:
hey mark, sounds like a neat project.
yeah, this is cool, and should be doable with compromise.

yeah - `toQuestion` is a little nutty. I remember that really doing my head in.
The good news is that we kind of conjugate participles already, it looks like we're missing [the "take" exception](https://github.com/spencermountain/compromise/blob/cd6910d3ac5a66cec53c30b77c4a2f094207837a/data/conjugations.js#L7).
this is how i'd do it - 
```js
let examples = [
  'eat a cookie',
  'take a shower',
  'watch the movie',
  'smell a flower',
  'join a club'
];

const toParticiple = function(str) {
  let doc = nlp(str);
  let verb = doc.verbs(0);
  let conjugations = verb.conjugate()[0];
  //use the Participle, or the PastTense
  let form = conjugations.Participle || conjugations.PastTense;
  doc = doc.replace(verb, form);
  return doc.text();
};

console.log(examples.map((toParticiple)));
```

(if you add `"taken"` to the data, then run `npm run pack`, the 'take'  example will work)

yeah, this is a little weird. will keep this issue open, so that it will become better

moberemk:
[...]

PR #569, opened 2019-02-28
Add participle form of "take"

moberemk:
Tangential to the issue in #568, this quickly adds the participle form of "take"


spencermountain:
cool, thanks. 
I'll take a look at why the CI's failing, too

Issue #583, opened 2019-04-20
Support for slashed verbs


elliot-nelson:
I've noticed that something funny happens with slashed (choice) verbs like "approve/deny" or "open/close".

For example: `nlp("The manager will approve/deny your request").debug()` returns the verbs I'd expect, but following up with `.verbs().toPresentTense().all().out()` returns the truncated string `The manager denies your request.`.  What I would hope to get back is `The manager approves/denies your request.`.

In contrast, `nlp("The manager will approve or deny your request").verbs().toPresentTense().all().out()` returns `The manager approves or denies your request.`, which is what I expect.

Perhaps a slashed verb could be treated as a (Verb SpecialConjunction Verb) phrase?

spencermountain:
hey Elliot, this is a great point. Thank you.

Yeah, the issue is in the verb-phrase grouping. Phrases like `would not walk` are grouped together as one verb, when they're conjugated and things. 
When there is a slash, they shouldn't be grouped together.
```js
nlp('approve/deny').verbs().length()
// 1
```
This should make the conjugation work properly, when fixed.

the verb-grouping stuff is [here](https://github.com/spencermountain/compromise/blob/master/src/subset/verbs/index.js#L101) if anyone wants to go for this.

charlesLoder:
I just discovered compromise and love the idea! I haven't contributed to open-source before so it may take me a bit, but I would to grab this if anyone isn't working on it.

spencermountain:
welcome @charlesLoder ! go for it!

yeah, this can be solved by adding some kind of check for a slash in that `find` method there.
Calling `.split()` in the right place there should do the trick just fine.

We don't have a clean way to search the whitespace-characters right now (i don't think!) - so I'd do something like this:
```js
let doc = nlp('approve/deny')

//each match
doc.list.forEach(ts => {
  //each term
  ts.terms.forEach(t => {
    //does it have a slash in the whitespace?
    if (t.whitespace.before.match('/')) {
      //pull it apart.
      doc = doc.splitOn(t.normal)
    }
  })
})
doc.debug()
```

¯\_(ツ)_/¯

I've written a first-pr thing [here](https://github.com/spencermountain/compromise/wiki/Contributing), if it's helpful.
cheers

charlesLoder:
awesome. I'll get to it

charlesLoder:
I added your suggested code:

```javascript
// src/subset/verbs/index.js

const find = function(r, n) {
[...]

PR #588, opened 2019-05-09
Issue #583 handle verbs w/ slashes

charlesLoder:
fixes: #583


scagood:
Would this compensate for adverbs?
Say for example: `I would/would not like that.`

Also, could you add test cases for this?

charlesLoder:
As is, no - it doesn't compensate for adverbs:
```javascript
let doc = `He hastily/speedily ran.`;
console.log(nlp(doc).adverbs().length);

>>> 1
```
Since `would` is a modal word, I'm not sure how compromise handles them. What I get is:
```javascript
let doc = `He would/would not like a puppy.`;
console.log(nlp(doc).verbs().out('array'));

>>> ['would', 'would', 'not like']
```

spencermountain:
👏

elliot-nelson:
Nice! :+1:

PR #596, opened 2019-06-25
unbreak typings and scaffold more detailed definitions

lostfictions:
hey there, looks like the typescript typings shipped with this repo have been broken for a year, since this change: https://github.com/spencermountain/compromise/blame/master/compromise.d.ts#L4

that renamed the namespace but not the corresponding function, meaning you can, say, get the `version` of compromise but not invoke the compromise function itself to do anything.

this is strictly worse than having no typings at all, since now typescript users have to do something like this for the compiler to not scream at them:
```typescript
import compromise from 'compromise';

;(compromise as any)('I went to the park').verbs();
```

this PR fixes that and makes the typings more detailed by scaffolding methods available on `Text`. it also fixes testing typings via [`dtslint`](https://hackernoon.com/testing-types-an-introduction-to-dtslint-b178f9b18ac8).

by the way, one quirk in the setup of this repo that might've made things harder if you were trying to set up a runner for typings in the past: `compromise` has a dependency of `compromse-plugin`, which has a dependency of... `compromise`. yarn will happily install compromise into compromise (i presume npm does the same) and this completely confuses typescript when you try to alias the typings, since it will find the older, broken typings in `node_modules` in addition to the ones in the local directory.

i used a relative import as a workaround and told `dtslint` to not complain about this -- but it's worth pointing out as a thing you might want to fix, since it will potentially cause weird stuff to happen (not just typings issues!) if you ever make a breaking change to the library.


spencermountain:
thank you!!
also, great eye with the circular dependency. I'll fix that today, then do a release.
🥇

spencermountain:
released as `11.14.0`

PR #613, opened 2019-08-13
Plural option in normalize is non-cooperative

rek:
Hey!! :)

Didn't see an issue for this, so making a placeholder here.

I was trying to use `plural` and `verbs` options together for normalize and found they don't work.

So I added this test. Then went looking in the code and also found:

`plurals: r => { //todo:this has a non-cooperative bug` (src/text/methods/normalize.js)

So looks like it is known about. I havn't had a chance to help looking into a fix yet though.

I also left this test as `.only`, was thinking to remove that when the fix is in this branch (and perhaps remove those two un-needed green tests).

### DIFF
--- a/test/unit/result/normalize.test.js
+++ b/test/unit/result/normalize.test.js
@@ -62,6 +62,45 @@ test('optional params', function(t) {
   t.end();
 });
 
+test.only('optional param - verbs and plurals together', function(t) {
+  var plurals = [
+    ['batmobiles', 'batmobile'],
+  ];
+  var verbs = [
+    ['I was walking', 'i walk'],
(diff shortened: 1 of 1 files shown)


spencermountain:
ah! Thank you Adam. ya sorry about that.
I've been working hard on [v12](https://medium.com/@spencermountain/compromise-in-2019-e7d0ca703320) for this exact reason. I'll merge this into that branch, to make sure this awful bug is finally resolved, and these tests get passed. I hope to release it in september, there's still a lot of work that's left.
thanks for this.

rek:
Awesome man, great work, love the lib.

PR #650, opened 2019-12-16
Fix file/directory name case

### LINKED ISSUE #649: Can't easily run tests on case-sensitive file-system
There are some mismatches between the cases of file/directory names and references to them in "require" statements.  This is often fine on macOS or Windows, but does not work with case-sensitive file-systems (like ext4 in Linux or inside WSL on Windows).

(I have a forthcoming PR with some simple `git mv` renames that should take care of it... I just want to be able to run all the tests successfully first!)

JaredReisinger:
The code references don't always match the actual filenames in case.  For example, `./src/index.js` refers to `./World/World`, but the actual path is `./world/World(.js)`.  This may work on macOS and Windows (which are case-insensitive by default), but not on Linux or on other case-sensitive systems.

This PR renames the files/directories based on what appears to be their canonical names (based on the source references and other directories).  Only two files (`scripts/pack.js` and `src/Doc/methods/output/_debug.js`) needed to be updated/edited to work with the new names.

Running `npm test` now succeeds 100% with no error messages on my system (WSL/ext4 on Windows), once I'd `npm install`'d in all of the plugin directories.

Fixes #649


JaredReisinger:
FWIW, `npm run testb`, and `npm run testOne` also pass 100%.  Sadly `npm run test:spec` fails, but that's because `tap-spec` isn't listed as a devDependency.  Hand-running that as `npx tape "./tests/**/*.test.js" | npx tap-spec` does succeed 100%.

spencermountain:
ahh, thank you Jared. Just got back from vacation, will look at this this afternoon.
❤️

spencermountain:
thank you!

PR #651, opened 2019-12-17
 Selectively ignore ".start" and ".end" matching token properties when handling a greedy match.

### LINKED ISSUE #648: start-anchor matching may have a bug (in v12)
Migrating our code to v12 (which is awesome, by the way!), and I'm running into a matching problem.  I've found a super-simple repro case, and will try to provide a PR if I can find the issue in the code... but I want to document it here first.

Take the example:
```javascript
const nlp = require('compromise');

nlp('a a b').match('a+ b$').found;   // returns true (expected)
nlp('a a b').match('^a+ b$').found;  // returns false (not expected!)  <=======
nlp('a a b').match('^a a b$').found; // returns true (expected)
```

The only difference is the addition of the start-anchor (`^`) requirement in the second match... and since the string/document _does_ start with 'a', I'd expect the match to succeed.  (Note that asking for 'a' twice explicitly seems to succeed as well... so it's a greedy/anchor issue, I think.)

JaredReisinger:
When handling a greedy match that's anchored to the start (`^`) or end (`$`), it is acceptable (normal, in fact!) for some of the matched terms to **not** be the first or last term in the phrase.  This fix attempts to allow that without accidentally forgetting that the greedy match as a whole should still either start or end the document.

Fixing this caused the `Doc.verbs()` behavior at https://github.com/spencermountain/compromise/blob/master/src/Subset/verbs/index.js#L15-L16 to change slightly (and correctly, I'd argue), such that a small change was needed at https://github.com/spencermountain/compromise/blob/master/src/Subset/verbs/methods.js#L45 (and on line 50) to compensate.

Added a test to specifically check for the anchored-greedy-match case.

(Note that this was fixed on-top-of/assuming PR #650, so that I could run the tests... I can isolate this fix if need be, but won't be able to verify it as well)

Fixes #648


JaredReisinger:
re: build failure... possibly because I did not `npm run build`?  I'm not sure at what stage builds are performed: pre- or post-commit/merge.

JaredReisinger:
Yay!  Do note that I did _not_ bump the version number in `package.json` since you may take other changes at the same time.  :-)

spencermountain:
wow! thank you so much for this. This is DEFINETLY the gnarliest part of the codebase, and you've done a terrific job. I really appreciate this.

Issue #654, opened 2019-12-18
Capture group doesn't work for .+ or *


Drache93:
Quick one since we're all ready for the holidays!
Happy to go into more detail when people are free.

Capture groups seem to capture everything BEFORE your target, if using `.+` or `*`. Using OR `(#Noun|#Verb)+` works fine.

Quick console example:

```
> nlp('John always ravenously eats his glue').match('john [.+] eats').tag('test').debug()
=====
  -----
  ｜ 'John'     - MaleName, FirstName, Person, ProperNoun, Singular, Noun, TitleCase, Test
  ｜ 'always'   - Adverb, Test
  ｜ 'ravenously'  - Adverb, Test
  ｜ 'eats'     - PresentTense, Verb, Test

> nlp('John always ravenously eats his glue').match('john [*] eats').tag('test').debug()
=====
  -----
  ｜ 'John'     - MaleName, FirstName, Person, ProperNoun, Singular, Noun, TitleCase, Test
  ｜ 'always'   - Adverb, Test
  ｜ 'ravenously'  - Adverb, Test
  ｜ 'eats'     - PresentTense, Verb, Test

> nlp('John always ravenously eats his glue').match('john [#Adverb+] eats').tag('test').debug()
=====
  -----
  ｜ 'always'   - Adverb, Test
  ｜ 'ravenously'  - Adverb, Test

> nlp('John always ravenously eats his glue').match('john [(#Noun|#Adverb)+] eats').tag('test').debug()
=====
  -----
  ｜ 'always'   - Adverb, Test
  ｜ 'ravenously'  - Adverb, Test
```

Last two examples are correct and expected behaviour of the first 2.

spencermountain:
ah, thank you.
Looking into the issue, i guess I see what [this comment](https://github.com/spencermountain/compromise/blob/dev/src/Phrase/match/03-tryMatch.js#L75) is for ;).

um, maybe I could ask your advice on something similar that I got stuck on - 
It may be the same issue:
```js
// === optional greedy at start ==
nlp(`wayne's world, party time`)
  .match('#Noun+? wayne')
  .debug()
// 'wayne's world, party time'

'aaa'.match(/b?a/)
//a


// ==== greedy-until-anything====
nlp('Toronto Ontario foo')
  .match('#Place+ .')
  .debug()
//'toronto ontario'

'aaab'.match(/a+b/)
```

I think your examples are the same as #2, right?
this stuff does my head in.
thanks

[...]

PR #655, opened 2019-12-19
Add support for named capture groups

Drache93:
First try. This only supports JSON matching and returns an object of terms when accessed:

```js
const match = [
    { word: 'dog', name: 'dog' },
    { tag: 'Verb', name: 'myVerb' }
]
const { dog, myVerb } = nlp('the dog played').match(match).named();

console.log(verb.text); // played
```

Could probably do with some help testing it against all the possible JSON inputs!

Also up for discussion, best way to handle multiple results for same match. Just pop them all in an array?


Drache93:
Example of using this to gather arguments for functions:

```js
  const templates = {
    equals: (a, b) => `${a.text}.status = '${b.text}'`,
  }

  const match = [
    { word: 'dog', name: 'equals-0' },
    { word: 'played', name: 'equals-1' },
  ]

  const res = nlp('the dog played')
    .match(match)
    .named()

  t.ok(res)

  const runnable = Object.keys(res).reduce((acc, k) => {
    const [fn, argIndex] = k.split('-')
    if (!fn || !argIndex) {
      return acc
    }

    if (!templates[fn]) {
      return acc
    }

    if (!acc[fn]) {
      acc[fn] = function() {
        const self = this[fn]
        const args = Object.keys(self.args)
          .sort((a, b) => a - b)
          .map(a => self.args[a])

        return templates[self.fn](...args)
      }

      acc[fn].fn = fn
      acc[fn].args = {}
    }

    acc[fn].args[argIndex] = res[k]

    return acc
  }, {})

  t.equal(runnable.equals(), "dog.status = 'played'")

  t.end()
```

spencermountain:
hey thanks for getting this started @Drache93 - got a couple thoughts.

I like `.named()`, but think it should return Doc objects, so you can do things like this:
```js
doc.match([{tag:'City', name:'from'}]).named('from').tag('Foo')
```
That would play nicely with the rest of the api, what do you think? 

I also like your idea for returning an object of named responses, and no 'null name'. That seems right to me, i think.

I think you probably found this already, but [here](https://github.com/spencermountain/compromise/blob/master/src/Doc/match/syntax.js) is the internal json syntax, once we've parsed the match string. I'm not sure it's perfect, but we should use it. I think you can pass its json into `.match()` and it will work already. We'll have to document it, if it's being surfaced to users.

adding the match name on `term.name` is clever - i'm concerned that's it's persistent though. It may be the wrong place for it. I ran into the same issue using `[foo]` capture groups, and if I can remember, the solution is awful. I think it involves matching a second time for each capture group.

Oh, i'm just looking at your #654 issue now, and it appears you've found that out, too.

I guess the capture group issue and the named-match issue are the same problem.

spencermountain:
just thinking out loud-

so a Doc object model looks like this:
```
Doc {
  Phrase { start, length }
  Phrase { start, length }
}
```
and if we're adding that a *part of a phrase* is a stateful thing, our Phrase object would need to remember more things, like which named parts are in which places..

```
Doc {
  Phrase { start, length, names:{from:{start, length}, to:{start, length}} }
  Phrase { start, length }
}
```
i dunno

Drache93:
Lot of thinking! Top to bottom...

* Thought you might say that about Doc, seems the norm for the library. I understood Doc/creating a Doc takes an array, which just means a little more work for lookup if I'm not wrong? But makes sense.
* Same page also with `named('test')`, makes sense for quick access, and chaining.
* Found the syntax, but I found console.log told me a lot more 🤣 - I'm a Typescript junky, I expect to be spoon fed :D
* `.name` definitely feels wrong, minimum should be added to Term class and supported in clone. You'll know more than me!
* Yeh I noticed a lot of comments in the code about the stuff in #654 needing to be worked on :D
* `doc.match('[#Place] to [#Place]').named('1')` also makes sense, or `doc.match('[#Place] to [#Place]').named(1)`. Personally I wanted to make sure I could control the order, but this seems like a good quick option, then full on JSON if you want more control. Not having to go into JSON is cleaner for sure; but always the down side of inventing syntaxes if using string. Like I previously mentioned I hacked `doc.match('[0: #Place] to [1: #Place]').named('1')` (again, that syntax was parsed into a match and tag heavy plugin - built in support so far is zero regex and 1 line :D )
* That's a fair solution, open's up the Phrase to doing more. Being able to refer back like that is great for highlighting in the UI what's being referred to

Passing in a variable for giving groups names could be an option `doc.match('[#Place] to [#Place]',[''from,'to'])`.

spencermountain:
i like that syntax - `[:name #Foo]` would work for me. I think (at least in the future) a capture group should be able to be multiple expressions, like `doc.match('captain of the [:team #Noun team]')`

also, maybe it should include *'all'*, or something like that, in the named responses? That way the whole thing isn't lost.

Anyways, I'm going away for the holidays next week and maybe it's probably best to punt this into the new year. Thanks for your help thinking through this!

Drache93:
Yeh that syntax feels familiar. Agree about multiples.  If `.named()` returns a Doc then won't need anything extra. Then just need a getter/transform?

So...

```
...named() = Doc
...named('name') = Phrase
...named(0) = Phrase (obviously only if a number has been used, so still accessing only by key) - in the future could be fancy and do `[: #Noun]` to get automatic numbering :P
...named().toRecord() = Record<string|number,Phrase>
```
Straight out of my head, so to be tweaked probably.

Have a good one @spencermountain! See you in 2020 🚀 🎆 🎉

Drache93:
Happy New Year @spencermountain!

Do you have any more thoughts for finalising the API and syntax for this?

Think we're agreed on the syntax `hello [:name #Foo]`. Saves the effort of documenting object based matching.

spencermountain:
happy new year to you too, Dominic.
Sure. Let me release 12.3, then we can start with something like this:
1) add `[:name ]` support to the syntax parser in `./src/Doc/match`
2) add support for `.names` as an object in the Phrase class
3) add `.names()` method to Doc
4) (hard part) properly support matching named-subsets in the `./Phrase/match/` match-logic.

I'm a little concerned about the 4th step, but I bet we can figure it out.

spencermountain:
oh hey, i just read about the [name capture group syntax](https://2ality.com/2017/05/regexp-named-capture-groups.html) coming to javascript - 

`[<name> #Foo]`

what do you think about that?

spencermountain:
oh, i think there's a question-mark too,
`[?<name> #Foo]`

Drache93:
Oh sweet, I missed that one. Yeh makes sense to just follow the spec. They show it with no space, so `[?<name>#Foo]`, but that would suggest this works: `[?<name>#Foo ?<desc>#Bar]` - which won't/doesn't/isn't correct usage in compromise I guess.

Maybe `?<name>[#Foo]` to avoid confusion about which part is named/captured?

PR #661, opened 2020-01-06
Feature/extend typings

Drache93:
WIP typing support for plugins.

Little messy, but attempting to re-use current type to allow chaining of `.extend`.

Note the use of `interface Compromise`. Basically, nlp being a type module stops us from doing smart stuff (i.e. module can't take generics). Also Document/World needs to be passed to everything else that needs them.


Drache93:
image
Result from extending a previously extended nlp!

spencermountain:
YESSSSSSS!
thank you. Let me know when I can merge this.

Drache93:
@spencermountain haha :D 
Fyi, the type will make it seem as though you have an nlp instance with only the plugins you've added, but of course that's not a thing and plugin is there even if you did an extend on a `new` nlp or the same one. Basically possible level of confusion and types not being in sync with reality...it's a start though!

Drache93:
Sexied it up a bit, typings now comes from plugin:

image

types/index.d.ts
```diff
@@ -329,8 +360,10 @@ declare module nlp {
     /**  */
     remove(): Document
     /**  */
-    hasOxfordComma(): Document
+    hasOxfordComma(): Document // Can we change all the unfinished types to 'any' instead? Not sure which ones are placeholders
```
Drache93: Fyi ^

Otherwise think this is good to go. Couldn't test plugins, so haven't updated their typings yet; how do you have the peer dep setup for testing them? npm link?

spencermountain: yeah, that's fine!
I don't have any good setup

spencermountain:
thank you for doing this!

types/index.d.ts
Drache93: Any idea which functions definitely return Document? I can swap the rest over to `any` then.

Build was failing for me btw `🇵🇷.` was coming out as `🇵🇷..`

text has post added to it in emoji tagger. And then again later Phrase text output (showPost). You building fine or should I make an issue?
```
  //test for unicode emojis
  if (term.text.match(emojiReg)) {
    term.tag('Emoji', 'unicode-emoji', world)
    term.text = raw
  }
```

spencermountain: yeah, that's the bug I have that's blocking 12.3.
I'll get it, Just need a little coffee first.

ah, this is cool. just playing around with it now.
I may rename some variables so this screen is a little less verbose, is that okay?
image

Yeah, almost everything returns `Document`, except for `.json()` methods and ngram methods. There will be a few exceptions.

Where should these files go? in respective `./plugins` folders?

Holy, this typescript stuff is HARD. Thank you for doing this.

Drache93: Haha it's very satisfying when it works! Can probably hide some of that inside a type. I'll take a look tomorrow.

Ah fair enough, feels like the post should be cleared if it gets added to the text - or at least a flag to know it was done.

Each individual plugin just needs to export a default nlp.Plugin, I'll do the numbers plugin tomorrow to show you

PR #662, opened 2020-01-08
Experiment with cleaner types, and pass world down

Drache93:
First look at cleaner extended types:

image


Drache93:
Document type now gets built from keys in extended type, so no parent data (i.e. nice and clean):

image

types/index.d.ts
```diff
@@ -39,328 +38,335 @@ declare module nlp {
...
+    join(str?: string): Document<W>
 
     // Output
     /**  return the document as text */
     text(options?: string | object): string
     /**  pull out desired metadata from the document */
     json(options?: string | object): any
     /** some named output formats */
-    out(format?: string): string
+    out(format?: 'grid' | 'color' | 'normal' | 'csv' | 'newlines' | 'html' | 'text'): string
```
Drache93: @spencermountain Would you mind double checking these? I did it awhile ago...

spencermountain: Sure, I'll update this on dev locally.

spencermountain:
Gonna merge in and take a crack at it while Britain is asleep.
THANK YOU 🤗

spencermountain:
hey @Drache93 I've added types for the other plugins, according to your newest format.
I also fixed that emoji-bug, so we can release 12.3.0 once you're ready. Lemme know if you're at a comfortable place with it, or would like another day.
thanks

Drache93:
Anywhere better to talk other than closed PRs? 🤣 

Looks all good, think I'm happy with it for now.
Only thing...

Am I right that the functions in plugins that return documents should return extended documents?...This could be a little weird since they should return document will all extensions, not just there own...
Also the plugins that add functions that return a class, should they return an extended document/basic one/or a reduced one?

iirc/if I understand correctly...the classes introduced in plugins reimplement document functions (i.e. they aren't using e.g. document implementation of forEach, they are implementing their own one)? In which case, they don't need to worry about other plugins add, the functions would just need to correctly return the class

Make sense? :D

spencermountain:
yeah, i understand. It is weird.
`doc.numbers().foo()` should return normal Doc objects, i think.
although I'm really not too sure.
Right now, at least, they do.

the classes created in the plugins are supersets of the `Doc` class - they don't usually re-implement forEach. They only times when they do, is when the method is overloaded, which is rare. `.json()` for example, gets overloaded frequently.

Hey, are you seeing this, too?
image

if I run `ts-node /types/types.test.ts` i get the same error. Is there something I can do?

Drache93:
Oh that's my bad, testing the main deps is now dependant on the numbers one working (so for me that was install and link). Probably not helpful since that's not automatic! Will include the types in the top level test.

Drache93:
So should `doc.numbers().numbers()` work?

spencermountain:
haha - oh boy. Never thought of that.
Yes it does. I just tried it.
But I can live with that ;)
deploying 12.3 to master - 🎈
I would not have been able to do this without your help.

Drache93:
Haha, will sort it out later then. Np! Thanks for the library :D

PR #663, opened 2020-01-09
Support named capture groups

Drache93:
Based on previous discussions. Looking at support that better involved Doc and Phrase, and returns Doc from basic function. Currently calling functions `.named()` to avoid overlap between Phrase function for easy access vs Phrase object. Seemed cleaner than putting everything into the Doc function.

- [x] Smaller code
- [ ] Performance review
- [x] Merge overlapping named groups into a single document
- [x] Support multiple groups per phrase
    - [x] Add support
    - [x] Make sure we're always giving terms a group name (or find a better way to do this)
- [x] Add syntax support for string based matching
    - [x] Choose string syntax
        - `[<name>#Foo]` ✅
        - `[<name> #Foo]` ❌ 
        - `[?<name> #Foo]`
        - `[?<name>#Foo]` - No space to match JS proposal `?<name>#Foo`
        - `?<name>[#Foo]` - Clarity that name effects whole group
        - `?<name>#Foo` - Make this entirely separate from capture groups
- [x] Add function for returning named groups as object
- [x] `Phrase.prototype.names` - object containg meta regarding named groups
- [x] `Phrase.named()` - returned named phrases uses `.names` object
- [x] `Phrase.named(target?: string)` -  returned target named phrases uses `.names` object
- [x] `Phrase.named(target?: number)` -  returned target named phrases uses `.names` object
- [x] `Doc.named()` - create document from named phrases
- [x] `Doc.named(target?: string)` - create document from target named phrases
- [x] `Doc.named(target?: number)` - create document from target named phrases
- [x] Store named capture group data in Phrase during matchAll
- [x] Persist `.name` across Phrases

Where else does support need to be added?

Support in tags object syntax checks if `capture` is a string, instead of boolean  this allows the rest of capture related logic to work as normal. Should we create a whole new value instead? i.e. `name`


spencermountain:
right! thank you.

`[<name> #Foo]` has my vote, it's simpler and would be a ~non-~(less-?)breaking change.

we could also support an optional question mark too, to observe the javascript standard.
`[?<name> #Foo]`
I guess we'd just throw the question mark away if it's there, what do you think?

Drache93:
I think capture groups are probably far enough away already from the standard it probably better to be different, avoid confusion/conflict. So no `?` I'd say.

Drache93:
Support added :D `the [<target> big dog] played'` and `the [<0> big dog] played'`

spencermountain:
yeah, i understand. That makes sense.
that means that this won't work, right?
```js
let doc = nlp('one two three')
let a = doc.match('one [<a>two]')
let b = doc.match('[<b>two] three')
```
that's fine, if so

Drache93:
You meaning the match, or the syntax? No space after `>` would simplify code, implemented it with a space after `>` (looking at it I think the space makes it more readable).

Since the named groups don't effect capture groups this should still work fine, also:

```js
let doc = nlp('one two three')
let a = doc.match('one [<a> two]')
let a2 = doc.match('one [<a> two]').named() // same as above
let a3= doc.match('one [two]') // same as above
let a4= doc.match('one [two]').named() // empty
let b = doc.match('[<b> two] three')
```

spencermountain:
oh ya, my concern was how things were stored on the Term object - if this were to introduce a consistency problem if a term was part of two 'name' results - maybe this is clearer?

```js
let doc = nlp('one two three')
let one = doc.match('one [<a>two]')
let two = doc.match('[<b>two] three')

one.named('a').text() //'two' ?
two.named('b').text() //'two'
```
I was just curious about that. Looking really great otherwise.

btw, if you have any opinions on the syntax, feel free to use your judgement. Yesterday i was leaning toward the colon one again, which now seems stupid. I could do a poll on the mailing-list, twitter if you wanted. Would that help?
cheers

Drache93:
I see your point, the actual data is tied to the Phrase like you previously suggested, just put on the term basically while that one match function is run and extracted afterwards, never looked at again. Probably worth deleting after it's used to avoid any confusion....or just pass the Phrase into the match function to be done directly??

I'm happy to try anything that is readable :D no harm in putting a poll out, there's probably pros and cons to using a syntax similar but not identical to the JavaScript proposal.

Drache93:
@spencermountain Another choice to make :D

How to handle overlapping names when outputting as object with `groupByNames()`:

* Replace (current - probably requires a warning)
* Keep old one, ignore new one
* Merge into array of Docs
* Merge into single Doc

Maybe supply an option in `groupByNames()` to select a method?...

spencermountain:
agh, i love this new feature. I'm starting to get excited about this.

I think I'd expect the result to be ***Merged into single Doc***.
ie: `res['type'].json()` has every result, from every phrase.

Drache93:
haha good to hear! Yeh that makes sense, then you can easily loop through the phrases.

Drache93:
@spencermountain So...just tested using no space after name `<target>#Noun` instead of `<target> #Noun`. Since it stops the name from being treated as a term, code becomes much simpler! Saving the name groups into the Phrase could still be simpler. But I think maybe any syntax without a space is better....less room for error (although merging the name term and the content term during the initial parsing could be a thing, but feels wrong)

Drache93:
Moved logic from 03-match into 03-tryMatch. Way simpler now.

spencermountain:
AHH this is so cool!!
I can't believe you've been able to do this.
🏆

Drache93:
Thanks! Probably worth writing more tests before release, but I think it's in a good place!

spencermountain:
hey @Drache93 I'm stuck on a capture group case, which may benefit from having your brain.
it involves multiple capture-group matches, per sentence:
```js
let doc = nlp('june the 5th, july the 7th, and sept the 12th.')

let m = doc.match('[<month>#Month]')
m.length //3 phrases  - all good

m.byName('month').length  // 9 - uh-oh
```

`p.names` is being set on all-3 phrases, so when we go to pull them out in `byName` we have too many.

just wanted to bounce this off you first, before I went down the match rabbit-hole.
Happy to fix it. Lemme know!

Drache93:
Hey @spencermountain, eek! Similar issue when trying to use named capture in post process so worth a better fix...

I used new ids for capture groups to ensure groups can have the same name but different start points....maybe would be better to not give them ids, just their name as key (also quicker lookup). Then store array of start/lengths instead of just 1 start/length per name? Almost like mini phrases inside of groups...

I can do it tomorrow (normal person time :D - my wife was being convinced today to move to Canada....) If you don't have time today

spencermountain:
Dominexit!
do it!
thanks. no rush

Issue #616, opened 2019-08-25
Many verbs at Past Tense are tagged as Noun


vroyibg:
For example "linked", "impoverished", "pounced" ...

spencermountain:
thanks Hải, I'll check it out.

spencermountain:
this should be fixed in `v13.0.0` - please let me know if you see any others.
cheers

PR #665, opened 2020-01-10
Dev

spencermountain:
##### 12.3.0

- prefer `@titleCase` instead of `#TitleCase` tag
- update dependencies
- fix case-sensitive paths
- fix greedy-start match condition regression #651
- fix single period sentence runtime error
- fix potentially-unsafe regexes
- improved tagging for '-ed' verbs (#616)
- improve support for auxilary-pastTense ('was lifted') verb-phrases
- more robust number-tagging regexes
- setup typescript types for plugins #661 (thanks @Drache93!)
- verb conjugation and tagger bugfixes


Drache93:
Is it worth a quick add of world functions or save for a patch? i.e. addWords etc.?

spencermountain:
hey @Drache93 good idea, let's do it in the next release

PR #670, opened 2020-01-18
Fix greedy capture

Drache93:
* [Issue #654)Capture `[*]`
* [Issue #654)Capture `[.+]`

Felt a little too easy...not sure I trust it!!


Drache93:
Just need to add named capture support, laptop died will get back to it!

spencermountain:
if the tests pass, it works!
This is terrific.

PR #674, opened 2020-01-20
Performance Tests

Drache93:
Playground for now


Drache93:
Need to compare properly, but pretty sure `performance.html` showed 4.7s before, now 3.5ish

spencermountain:
i think that's a terrific idea, and a really cool optimization.
I've not kept-up with babel/es6 current best-practices - do we need a polyfill for Set/entries?
Yeah, sorry for making a mess on dev yesterday. I was in the zone.
I'm happy to go through and cleanup the notA data. It's pretty ad-hoc.
Also, that benchmark thing looks amazing.
Let me know if I can help!

Drache93:
Haha, was great to see all the changes on Dev, some really cool stuff you're working on there.

I'll definitely need your help for cleaning up any tags, I imagine it helps if you know what tags should and shouldn't be used for each thing. But I'm mostly playing around atm, so don't worry if you have other things! How close is v13? Any priorities?

As far as I understand `@babel/preset-env` should cover it. Come to think of it, most of your code is already ES6, e.g. arrow functions, let, classes. But looking at the build, Set and entries are still there...
I'm not obsessed with the idea of Set, but it does shortcut indexOf and duplicates in the same way an object does...however you can't add multiple values at once, so still need a loop for that...

I'm tempted to store tags in a class instead to give us some way to add/remove more efficiently.

Drache93:
1 observation, performance is basically linear, so a lot of the time taken is just because of how much is being done. I think smarter ways to NOT do things is probably the biggest place to make gains...

spencermountain:
yeah, I agree.
when you thought of this idea, i thought about making some subset of `./corrections` that was less-picky, or less-safe - kind of a deterministic overwrite of the tags, or something. 
... `.tagFast('Person')` could just set `term.tags = personTags`, and no other fuss.
Eventually a term's tags get too delicate to do that, but maybe in some cases we can afford to do this.

Let's try to release 13 this week. I don't have a specific list of anything. does that work for you?

spencermountain:
I went back and forth on trying to cache Doc objects this for like a month, and don't really understand how to measure performance very well. I bet you'd be able to find some clever things to do there, too.

The nice thing about `./corrections` is that only the tags ever change, so theoretically everything can be cached way-up high - then speed through the various `.match()` statements much faster. I tried doing this on `Phrase` and `Doc` - I would call a `.freeze()` method before corrections runs, then `.unfreeze()` afterwards. I assumed it would be amazing but then it wasn't. The parent-child stuff was too weird. I couldn't figure it out.


there's also a ton of matching that can be re-written with `<name>` now. That may really help, too.
ok, end of brain-dump

Drache93:
That might make sense. The fact that things get delicate is maybe something to look at :D
Yeh that's fair, at this point I'm just looking for work to do, everything's look pretty good! So probably no real PRs for now. 

So far I did the simplest way I know (Date.now) to check performance plus I did a `nlp.clone()` to try and avoid too much caching between tests. Other than that, same method as your performance.html. I think anything more complex probably not worth the effort, as long as we can see that our changes are having an effect, the metric itself is probably less important. I'm mostly using the average time difference, there's some wiggle between runs, but very obvious when things go right/wrong.
First time I've properly looked at performance recordings in chrom?ium and it's really useful. Can already see that unTag is further down the list now.

Hmm that's fair, considering it's behind the scenes, would make sense to have internal functions to skip a lot of the normal steps. Was surprised to see corrections is literally just match/tag :D

Another note, I wonder if bitwise operators might be a smart way to get around the complexities of playing with arrays...

Drache93:
Funny enough, creating a tagUnsafe that basically skipped all logic in tag for Doc and Term, and set the value directly on the term object...didn't result in any reliable change in perf (just replaced all .tag - not tagSafe - in fixPerson).

spencermountain:
hey, what do you think about the following changes?
* change `'capture'`  prop to `'name'` in syntax obj
* combine `named()` and `.groupByNames` into one function `byName`?
```js
let m = nlp('fifth of november blah').match('[<date>#Value of #Month]')

m.byName('date').text() // 'fifth of november'
m.byName('fooo').text() // ''

m.byName() // { date: Doc }

nlp('asdf').byName() // {}
```
I can alias the current methods, if you'd like.

spencermountain:
also, and this one is more aggressive:
* stop returning [capture-group] by default?
```js
let doc = nlp('5th of June')
let m = doc.match('[#Value] of #Month')

m.text() //'5th of June'   (whole thing)
m.byName('0').text() //'5th' 
```
that would be a pretty-big change, but it makes sense to me, especially when there's multiple capture parts - it's not clear what `m` is going to be.

what do you think? I'll wait before changing anything, to see how you feel

Drache93:
No complaints. Returning a Doc containing only named is probably useless since you still need to work on it to get them, personally I only need the object response. So yeh, no complaints to replace/remove!

So was playing around with minor fixes and places to make shortcuts. But thinking about your point of conceptual wins and thinking maybe better targeting something more specifically...(brain dump)

So...corrections are a core part in getting the output we currently have. So feels like they should be involved as the base data/lexicon. So maybe an argument to include it in a more advanced way also.
When it comes to processing though, I'm wondering about the idea of "Batching matches". This is made easier by the fact that the syntax is converted to object already, but basically we merge a bunch of matches to remove overlap and result in us only running match once. The named syntax makes it easier to handle the output...
Passing in an object of name => tag mapping would allow us to pre-prepare the tags, again removing overlap, and including lineage....
Came to me a few minutes ago :D so thoughts?

Drache93:
On your second post => using number to access groups in order feels natural (fyi number is supported as well as string!). The current capture group does feel a bit like you're losing too much, when you sort of want the capture group for later/other stuff...?

Question is, how useful is `m` afterwards we've captured. And, since chaining is nice, do we want a way to return to the root after we've tagged the capture group? Was trying to find a Gremlin/Tinkerpop example...

spencermountain:
cool!
that's a really neat idea. I've never thought of that....
would it still have to check every phrase, for every match?

ya, I heartily welcome any sort of structuring of the stuff in ./corrections. It's been a mess for a long time. I've always wanted to see how many statements are actually redundant, or just useless.

re:chaining => Yeah, good point i can look at testing `byName().parent()`.

let me know when this pr is ready, i can't wait to use that new performance metric.

Ya, this named match stuff is pretty hot. It kida changes everything, and makes compromise  less-awkward to use, doesn't it. 
🏅

Drache93:
Still thinking... they'll still probably be multiple matches, but the idea is by knowing ahead of time what is intended (e.g. tags) we can cut down on duplication I think.

Yeh fair point, they're so specific though they'll easily be unknown side effects...but might be better off pushing some of those unique cases to the users to handle themselves.

I'll cleanup tomorrow, was going to keep it for research but yeh makes sense to have the metrics in.

Haha yeh, definitely satisfying being able to grab input so easily

Drache93:
@spencermountain This is ready!

spencermountain:
thanks!

PR #677, opened 2020-01-22
Fix duplicate named groups

Drache93:
Not fixed yet, just made the PR so you're aware.

Since named groups are given a unique id, seeing duplicates when used in a plugin where we do a named match as part of a postProcess - see `named-match.test.js`

Might be an issue of clone not using fresh Doc?
Post process runs over my input 3 times (correct Doc), even though I'm only using the plugin twice.


Drache93:
Feels like I'm miss using post process? Fixed by doing this all is Doc.prototype instead.

PR #679, opened 2020-01-23
First-pass GitHub Actions workflow for building/testing compromise

JaredReisinger:
This is an attempt at a very basic build/test workflow for GitHub Actions that can be triggered on any push (to master, or any branches or PRs).  The set of tests run may be excessive, as it wasn't clear to me what the difference between the various tests listed in `scripts` in `pacakge.json` meant.

In addtion, some missing devDependencies (eslint, nlp-corpus, nyc, tap-spec, ts-node, and typescript) have been added to ensure that the tests can all run (even if they aren't all passing at the moment).

Finally, a new `scripts/plugins.js` was added to perform arbitrary commands for each plugin.  The hope is that `scripts/plugin-check.js` can eventually be deprecated and replaced by (for example) `npm run plugins:build`.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -41,13 +41,13 @@ jobs:
           npm run build
           npm run plugins:build
           npm run test:smoke
-          npm run test
-          npm run test:spec
+          # -- currently failing -- npm run test
+          # -- currently failing -- npm run test:spec
           npm run test:perf
           npm run test:types
           npm run test:stress
-          npm run test:speed
+          # -- currently failing, also alters on-disk file -- npm run test:speed
           npm run plugins:test
-          npx nyc tape ./tests/**/*.test.js
+          # -- currently failing -- npx nyc tape ./tests/**/*.test.js
         env:
           CI: true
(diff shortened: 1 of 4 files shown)


JaredReisinger:
Hey, @spencermountain ... just found out that the GitHub actions **will** run in my fork.  Let me vet them there before you merge this!  (I'll remove the [WIP] tag and comment once it's actually ready...)

JaredReisinger:
Okay.... here's where this stands: at present, the workflow passes as long as `npm run test`, `npm run test:spec`, `npm run test:speed`, and `npx nyc tape ./tests/**/*.test.js` are all commented out.  I believe that all of these _other than_ `test:speed` are failing from the same underlying test.  The speed test appears to (a) expect to be run on node 10.x, and (b) modifies a file that's not .gitignore'd (to persist the last-known-speed, I think).

With the caveat that there may be changes in dev that need testing tweaks before the basic tests can be included, I think this is "runnable" by default in the compromise repo.  At the very least, it should be possible to include this in parallel with the Travis CI pipeline that already in place.  Because of the way GitHub Actions work, merely including `.github/workflows/...` files will cause them to get triggered and run.

spencermountain:
YESSSSSSS
thank you Jared, this is terrific.

PR #680, opened 2020-01-24
Fix speed test and types

Drache93:


Drache93:
Ah, looks like PRs don't get done by Actions automatically?...

spencermountain:
ya thanks - 
I love your idea of storing past performance, to catch performance issues. 

I guess it runs a differently on a supercomputer, vs a laptop, but maybe a greater variation is using different node versions. Node 6->Node 13 has a huge effect on speed.
would it make sense to benchmark some node versions, in that performance file? 
If compromise gets slower or faster, i wouldn't mind editing that file myself. That would be fine.
I'm so glad your brain is able to do typescript. I see a `<>` and my ears start ringing.

spencermountain:
hey, do you have a preference for gitter, or slack?
i've setup both in a sloppy way [slack](https://slackin-kyzvclgjlg.now.sh/) [gitter](https://gitter.im/nlpcompromise)
and have people chatting on both. it's a mess. Wanna join the gitter room? I'll get Jared to join that, too. I'm gonna try to get this working today, but it's already noon:
```js
nlp('one [two] three').text() //one two three

nlp('one [two] three').byName(0).text() //two
```
that's (maybe) the last feature for v13?

Drache93:
Hahaha I've found I enjoy the art of twisting the types to do things that would never be allowed in a normal language :D
Totally agree speeds will be wildly different everywhere, so I think simply logging it and being aware of it is a good place to be. Fact is, locally if we are doing performance focused changes, we can easily test our changes - but no point obsessing over it at all levels for sure.

Don't think I had heard of gitter until just now, onwards ->>

PR #681, opened 2020-01-24
Add term types

Drache93:


Drache93:
Happy to do this properly (full extend support) in a minor/patch instead of 13 if you prefer @spencermountain

spencermountain:
this look great. you are amazing.

PR #683, opened 2020-01-24
Fix

Drache93:
So index is only used if no name and named groups are always made for capture groups. Probably doesn't need to be called "named" anymore!

Function might be nice to return the array of non-named capture groups


spencermountain:
🏆

PR #685, opened 2020-01-27
Fix capture of named OR groups

Drache93:
Clean up reg group with only name in it


Drache93:
@spencermountain Feels a bit hacky, do you think a change to the syntax parsing would be better?

spencermountain:
good find!
thank you.

PR #689, opened 2020-01-29
Hotfix/named overlap

Drache93:
Since named groups are now basically default for capture groups, I renamed the "names" item on Phrase to "groups", was called groups during all the matching, then renamed by the time it got to Phrase and document!

Can undo if you prefer names/named - I removed the old alias' for byName and added a `groups` alias


Drache93:
So..."Phrase.has()" doesn't return phrases, so no capture groups assigned due to these fixes (was working accidentally before because we were always adding to the static `names` value)...also it returns immediately after one result found, we would need to go over every phrase to build the groups...which would slow to has() dramatically.

Tempted to call it a feature considering `Doc.if` and `Phrase.has` aren't matching functions and return the input phrases unchanged...
Is it worth having a flag for if/has that ensures groups are built? (don't matchOne and update `this.groups` with results)

Fyi, the fix in the PR is different from what I mentioned yesterday, I realised the issue here was we weren't just applying the groups to the matches/Phrases they came from.

PR #692, opened 2020-01-30
Release/13

Drache93:
Just need to update tests etc....

### DIFF
--- a/plugins/dates/src/01-tag/00-basic.js
+++ b/plugins/dates/src/01-tag/00-basic.js
@@ -118,7 +118,7 @@ const fixDates = function(doc) {
       .tag('Time', 'date-before-Cardinal')
     //saturday am
     date
-      .match('#Date [(am|pm)]')
+      .match('#Date [(am|pm)]', 0)
       .unTag('Verb')
       .unTag('Copula')
       .tag('Time', 'date-am')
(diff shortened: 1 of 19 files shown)


spencermountain:
i spent a long time staring at tryMatch last night and couldn't figure that out.
Nicely. done.
will release v13 today.
🍺🍺

spencermountain:
hey, i'm seeing some `undefined` group names, when they are unnamed
```js
let doc = nlp('one two three four')
let m = doc.match('one [two] three [four]')
console.log(m.groups())
//{undefined: Doc}
```
want me to take a look?
added a couple tests in `./match/named-silent.test.js`

Drache93:
Ah right we're not falling back to using the index if no name given (when generating the output object). 

Few questions:
* Do we want to return indexed items with groups, or just the named ones, should there be an option?
* Do we want a function for just returning the indexed items in the right order?
* Should we be able to grab an items using `.groups(myname)` and get it based on index OR name, currently index is ignored if name exists...

Not sure if mixing index/named is too much for one function. No harm though in just returning everything - so easy fix for above is just `group || index`

Drache93:
Also...we support looking up group with `.group(0) OR .group('0')` once we're forced to create an object, we have to choose one for the index...can't have both unless we store twice
So maybe an argument there for groups only returning named groups, and something else for returning array of indexed ones?....

spencermountain:
yeah, i like the softness about supporting both numbers and string-numbers. That's great.
I don't mind doing `.groups()['0']` or `.groups()[0]`. Javascript just mungles them all together, anyway:

image

i hate thinking about this, but apparently javascript arrays are actually just objects, and even run number-indexes through a hash, before looking them up. Urgh. 

we should build a rust fork. 😞
lemme know if this is a hard thing to fix, we could ship v13 as is.

Drache93:
Yeh my bad, you're right they both work! By brain runs ahead of me at times :D
I won't argue with rust! PR ->

PR #693, opened 2020-01-30
Some fixes

Drache93:
So current problems:
* undefined group name - fixed
* Not incrementing group number (didn't check for true) - fixed

Last one breaks some two-word capture, because groups aren't being merged together...without it `[a] test [b]` gets merged into one, which it shouldn't.

Basically, need another look at postProcess in Doc/syntax as we want to ensure we fill in gaps for a named capture group, but don't join unnamed groups together.
Might be worth giving unnamed group a named during parseToken, with a group counter in syntax...
I might not have time to look more until tonight.


Drache93:
@spencermountain So we happy to ignore side by side groups no longer being merged?

spencermountain:
that seems okay to me, what do you think?

Drache93:
Probably makes more sense that each group is different - weird having them merge automatically now that they're named!

PR #695, opened 2020-01-31
Add a groupType to match tokens so we know start/end

Drache93:


spencermountain:
this is terrific! thank you for doing this so quickly.

spencermountain:
image
🎉

PR #709, opened 2020-02-11
Make sure extended types follow you everywhere...

Drache93:
Basically makes sure you can use all functions normally from within your plugin and with your extended types. Plus also inside loop functions.


spencermountain:
just in time for 13.1.0!
thank you!

Drache93:
Good timing!

PR #714, opened 2020-02-20
Add "release" workflow

JaredReisinger:
This is a very simple workflow to attempt to automate publishing releases to npm.  I was hoping to drive the version number in `package.json` from the GitHub release, but `npm version` would attempt to create a git tag which would conflict with the one created directly by the GitHub release process.

**Requirement:** You must have updated the intended version number in `package.json` _before_ you create the matching release in GitHub.  This will ensure that `package.json` tagged in release `X.Y.Z` says `"version": "X.Y.Z"`, and that version will also be used when publishing to npm.

**TO DO:** Four additional steps need to happen to make this work:
1. You'll need to create an npm authentication token via https://www.npmjs.com/settings/spencermountain/tokens.
2. You'll need to add this token as a secret named `NPM_TOKEN` at https://github.com/spencermountain/compromise/settings/secrets.
3. If you currently require 2FA to publish (at https://www.npmjs.com/package/compromise/access), you'll need to turn it off.
4. Once you can verify that it's doing roughly the right thing, the `--dry-run` flag should be removed from the `npm publish` command.

(The 'build-and-test' workflow was tweaked to make it easier to find failing steps... instead of having to sift through ten separate npm commands, they are bucketed into rough groups: install, lint, build, and test; each one of which is a small number of underlying npm commands.)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -3,8 +3,11 @@ name: Build and test
 on: [pull_request]
 
 jobs:
-  build:
+  build-and-test:
     runs-on: ubuntu-latest
+
+    env:
+      CI: true
 
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


JaredReisinger:
Gah... I was worried about this... having the `.npmrc` file use `NPM_TOKEN` takes over way too often, breaking most `npm` commands for anyone who's even cloned the repo.  I'm going to see if I can get "just-in-time" auth instead.

JaredReisinger:
Better.... the "normal" process is good at any rate... you'll have to test out the release process yourself, but you can use a dummy release (not a version number) to at least verify that (a) the workflow runs, and (b) it gets as far as the dry-run publish.

spencermountain:
hey cool!
I don't mind setting the package number explicitly.

so instead of running `npm publish`, i click on this bad-boy, and it publishes automatically?
image
that works for me!

spencermountain:
hey this dry run worked! 
what a cool thing. Thank you so much for doing this.
removed the dry-run flag, will trigger it next release.
thank you!

PR #757, opened 2020-05-28
Fixing Dependencies

MarkHerhold:
Dev dependencies are marked as normal ones. Please double check to be sure I got the fix correct!


spencermountain:
ah! thank you.

PR #766, opened 2020-06-27
Improvement/add name (Vladimir Putin)

fmacpro:
- added "vladimir" to maleNames.js
- added "putin" to lastnames.js


spencermountain:
thanks

PR #772, opened 2020-08-06
prevent doc offset changes (length) from clobbering term[0] offset

### LINKED ISSUE #771: offset information for term[0] is clobbered by offset information for the Doc as a whole
_(Using v13.3.1)_

```javascript
> nlp('hello world').json({ offset: true })

[{
  text: 'hello world',
  terms: [
    {
      text: 'hello',
      tags: [Array],
      id: 'hello-QHp27op',
      pre: '',
      post: ' ',
      offset: { index: 0, start: 0, length: 11 }  // <====== notice the length, it's the same as the doc's length below
    },
    {
      text: 'world',
      tags: [Array],
      id: 'world-GYTaM1w',
      pre: '',
      post: '',
      offset: { index: 1, start: 6, length: 5 }
    }
  ],
  offset: { index: 0, start: 0, length: 11 }  //<====== same as term[0]!
}]
```

The problem stems from https://github.com/spencermountain/compromise/blob/master/src/Doc/methods/output/_offset.js#L42-L43, where the doc-level offset infomation is "assigned from" term[0].  This isn't a copy of the data, it's the same underlying object.  Thus, when line 43 sets the length for the Doc's offset, it is _also_ setting the length on term[0]!

I have a fix locally that uses `Object.assign()` to create a new Doc-specific offset object.

JaredReisinger:
The original code assigned to the doc-level '.offset' result from the offset information from term[0].  This creates an additional reference to the same underlying object, *not* a clone of the data. When the doc-level length was then set, it clobbered the length of the term[0] offset info.

This now uses `Object.assign()` to create an all-new doc-level offset object with the same values as it had previously.

Fixes #771

### DIFF
--- a/src/Doc/methods/output/_offset.js
+++ b/src/Doc/methods/output/_offset.js
@@ -39,8 +39,16 @@ const calcOffset = function(doc, result, options) {
       //   console.log(t.post)
       //   return n
       // }, 0)
-      o.offset = o.terms[0].offset
-      o.offset.length = o.text.length
+
+      // offset information for the entire doc starts at the first term, and
+      // is as long as the whole text (note that there may be an issue where
+      // leading punctuation is counted in the doc text length, but is
+      // *excluded* from the term[0] start position)
+      o.offset = Object.assign(
+        {},
+        o.terms[0].offset,
+        { length: o.text.length }
+      )
     })
   }
 }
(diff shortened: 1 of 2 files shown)


spencermountain:
you are the best.

PR #775, opened 2020-08-07
Fix doc/result offset start

### LINKED ISSUE #774: Doc offset uses term[0] start position, but might include terms[0].pre at the beginning, and "length" includes these characters
Found this while investigating #771, but didn't get it "snuck in" before that one was merged.  😉 

```javascript
> nlp('"hello world"').json({ offset: true })

[{
  text: '"hello world"',
  offset: { index: 0, start: 1, length: 13 },  // <====== takes 'start' from term[0], but "text" includes leading punctuation (aka terms[0].pre)

  terms: [
    {
      text: 'hello',
      tags: [Array],
      id: 'hello-KpoKa0W',
      pre: '"',                                // <====== this "pre" value is included in the doc text
      post: ' ',
      offset: { index: 0, start: 1, length: 5 }
    },
    {
      text: 'world',
      tags: [Array],
      id: 'world-sLkFApK',
      pre: '',
      post: '"',
      offset: { index: 1, start: 7, length: 5 }
    }
  ]
}]
```

I think the right thing to do is subtract the relative position of term[0] from the start of "text"... `text.indexOf(terms[0].pre)` would be "1" in this case, indicating that "text" starts 1 position before terms[0] really does.  By shifting the main offset to `{ start: 0 }`, the length now makes more sense.

(PR to come)

JaredReisinger:
In cases where leading punctuation is included in a result/doc, we need to account for that difference from the `terms[0].offset.start` position.  The easiest way to do this is simply to subtract any leading positions by looking for how far into the result's `text` that `terms[0].text` starts.

Fixes #774

### DIFF
--- a/src/Doc/methods/output/_offset.js
+++ b/src/Doc/methods/output/_offset.js
@@ -40,15 +40,16 @@ const calcOffset = function(doc, result, options) {
       //   return n
       // }, 0)
 
-      // offset information for the entire doc starts at the first term, and
-      // is as long as the whole text (note that there may be an issue where
-      // leading punctuation is counted in the doc text length, but is
-      // *excluded* from the term[0] start position)
-      o.offset = Object.assign(
-        {},
-        o.terms[0].offset,
-        { length: o.text.length }
-      )
+      // The offset information for the entire doc starts at (or just before)
+      // the first term, and is as long as the whole text.  The code originally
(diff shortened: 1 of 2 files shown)


spencermountain:
ah thank you.
let me know if this is a rush - otherwise i'll release it later this week.
cheers!

PR #781, opened 2020-09-25
🕹️ Updated Sample Code Snippet

the-mcnaveen:
From: https://github.com/nlp-compromise/nlp-pronounce/issues/1#issuecomment-699048674


spencermountain:
thank you!

Issue #765, opened 2020-06-27
new match2 plugin


kelvinhammond:
Hello,

I'm building my own match function, probably add it as an extension because it requires a dependency which will increase the code size.

I have so far figured out that I can use `termList` to get all the terms to match against and now I'm trying to save the groups. How do I set the groups for a matched term, I need to set multiple by name, number, and the total match as group `0`? The code for the `match` function is a bit hard to follow.

Also turns out using a subset of `termList` with `buildFrom` to build a doc doesn't actually work as expected, it fails when I run `text()` on the new document.

spencermountain:
hey @kelvinhammond that sounds really neat. I'd love to hear more about it.
Yeah, the internal-api is a bit complicated. The groups concept especially so. You're right that `buildFrom` is the method I use to compose new Text objects, with a shared context and data.

the best way to get a feel for groups and buildfrom may be to place a console.log [here](https://github.com/spencermountain/compromise/blob/master/src/Phrase/Phrase.js). Groups is an object - I think numbers will also work as the keys, like you mentioned.
yeah, if it is not set correctly, the linked-list will blow-up when you call `.text()`. Crawling that is the most expensive operation we have, so it's not possible to validate it at create.
Can help further if you have more questions
cheers

kelvinhammond:
@spencermountain 

So far I've:
1. Used [chevrotain](https://sap.github.io/chevrotain/docs/) as the parser for the regular expression. This is the library which I've imported which would increase build size.
2. Implemented pike's vm similar to [this][pikevm] but on words.
3. Implemented groups with subgroup support
4. Match zero or one, zero or more, and one or more
5. Greedy / non greedy matching

I still need to implement:
- proper saving of matches along with their groups, excluding non-capturing groups.
- range modifiers
- end of `$`
- allow matching through string and not just at start of while excluding that match, default is start of for the machine
- support regex matches

When it comes to compromise currently in my scratch I'm doing:
```javascript
let ast = null;
// parses the regex and set the ast.prog to a "program" which runs in the vm
ast = parse("remind (you|me|to)+");  
const doc = nlp('remind me to work on a better match');
// matches = { found: true, saved: [terms] }. this will change
const matches = pikevm(ast.prog, doc.list[0].terms());
console.log(matches?.saved?.map(x => x.text));
```
outputs:
```javascript
[ 'remind', 'me', 'to' ]
```

[...]

PR #796, opened 2020-11-14
Add match2 with history to plugins directory.

kelvinhammond:
Related to https://github.com/spencermountain/compromise/issues/765


spencermountain:
hey this is great! Nice work.
Gonna merge it to dev and take it for a spin. Will ping you if I change anything non-cosmetic.
What a cool plugin!!

spencermountain:
hey @kelvinhammond  what do you think? any feedback is welcome
https://github.com/spencermountain/compromise/tree/master/plugins/strict
I learned a lot going through this - thank you for doing it.

kelvinhammond:
@spencermountain The README can be updated to say that it supports.

- [Method match functions](https://github.com/spencermountain/compromise/blob/master/plugins/strict/src/pikevm.js#L198)
  - [see format here](https://github.com/spencermountain/compromise/blob/master/plugins/strict/src/parser.js#L31)

Example: `(@hasContraction|@isTitleCase)+`

I forget when I added this but its supported and should be added to the README.

Looks good to me, thanks for adding it to the project. You can call it whatever you'd like. Why `strict`?

spencermountain:
added! Thanks, yeah I was thinking it was a match-parser that would throw an error if it the input was wrong.
`compromise-formal` was the 2nd choice. Happy to change it to whatever.
Happy holidays.

kelvinhammond:
Its a completely new match parser with its own syntax and features. I tried to implement what the current `match` supports but with the syntax being closer to an actual regex and supporting nested values and a few other issues with the original `match` method. 

In summary, it is a bit more strict but its also completely new with its own features. If its missing anything from the original `match`I may be able to implement them depending on the feature.

Happy holidays to you too.

thegoatherder:
@spencermountain is this plugin still alive or dead? Doesn't seem to be in the plugins dir any more...

kelvinhammond:
@thegoatherder @spencermountain I think it was removed, I could help maintain it if you'd like.

Issue #811, opened 2021-02-06
TypeScript errors after updating to latest version (v13.9.1)


ihorskyi:
According to release log feature **add fuzzy option to match commands** was introduced in v13.9.0
Current type definition for `match()`is:

```typescript
match(match: string | ExtendedDocument<Ext, W, Ph>, options:any): ExtendedDocument<Ext, W, Ph>
```
That means that `options` is required param now, thus all my `match, matchOne, etc...` methods have errors.

<img width="962" alt="Screen Shot 2021-02-06 at 6 16 13 PM" src="https://user-images.githubusercontent.com/6171149/107131766-75327400-68a7-11eb-939d-a67c0c54874d.png">

Looks like `options` param is optional (some pun intended), so we should have it this way:

```typescript
options?:any 
vs 
options:any 
```

I created PR to fix this and few typos along the way

spencermountain:
released as `13.9.2`
thank you for the heads-up Ihor!

PR #812, opened 2021-02-06
Fix typings for match options to be optional

ihorskyi:
Fix for #811  (TypeScript errors after updating to latest version (v13.9.1))


spencermountain:
yikes! thank you!!

PR #838, opened 2021-04-28
Fix broken table

terrierscript:
I found broken readme table.
Sorry but I'm not sure that's the correct classification (tag -> concept / split -> API)


spencermountain:
yikes! thank you @terrierscript !

PR #847, opened 2021-06-18
Fix tokenization out-of-memory crash and slow performance for long string, few sentence case

srubin:
We ran into a case where `compromise` crashed with an out-of-memory error. Here's a simple repro case:

```js
test('tokenize() does not crash on long string with few sentences', function (t) {
  let text = 'The quick brown fox jumped over the lazy dog.\n'
  text += '--\n'.repeat(100000)
  let doc = nlp.tokenize(text)
  t.end()
})
```

The issue is that the `chunks` array in `splitSentences` was growing in memory roughly quadratically as it processed each non-sentence chunk. The easy fix to this is to clear each chunk after processing it (i.e., move `chunks[i] = ''` out of the else block.

This fixes the OOM crash, but the performance of `splitSentences` is bad. A trace showed that we spent 50 seconds in `isSentence` in the above test case (because it's testing each of the regexes on the entire combined string, i.e., `--\n` then `--\n--\n` then `--\n--\n`, etc.). To fix that, this PR refactors the checks in `isSentence` to include early exit conditions. The total time to run that test goes from 50 seconds to ~100ms.

The code with passing the context (`prefixContext: { hasLetter: boolean }`) into the function isn't the cleanest thing ever, but that added complexity is at least all local to a single file.

### DIFF
--- a/src/01-tokenizer/01-sentences.js
+++ b/src/01-tokenizer/01-sentences.js
@@ -40,7 +40,7 @@ const testIsAcronym = function (str, suffix) {
   if (suffix.indexOf('.') === -1) {
     return false
   }
-  return isAcronym.test(str);
+  return isAcronym.test(str)
 }
 
 const testHasEllipse = function (str, suffix) {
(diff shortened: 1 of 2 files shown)


cvanwinkle (approved):

spencermountain:
this is beautiful. Thank you so much.
Let me do a release this afternoon.

spencermountain:
hey @srubin this is released now as `13.11.3` 
nice solution, for a really tricky bug!

also, Descript seems like a really neat product.
cheers!

srubin:
Thanks for the quick release!

Issue #787, opened 2020-10-12
Offset after clone


spencermountain:
offset requires mapping a match back to the parent doc, but .clone breaks the link between them.

```js
let doc = nlp(`so I have not arranged`)
doc = doc.clone()
console.log(doc.json({ offset: true, normal: true })[0].offset)
// { index: undefined, start: NaN, length: 22 }
```

PR #874, opened 2021-09-08
attempt to provide plugin-level options for compromise-dates

JaredReisinger:
_**NOTE:** This might very much **not** be the way you'd want to approach something like this... but it served a need in my particular case, so I wanted to offer it up in a PR._

I have a use-case that needs `numbers().normalize()` _**not**_ to run globally just because we're attempting to tag dates in the input.  Why?  because `numbers().normalize()` will mutate the input text/doc by injecting a space character between numbers and suffix text, changing `123abc` into `123 abc`.  We perform post-processing on the parsed result, using it to augment the original input string, so it is _absolutely critical_ for us that the parsed information matches the original input when it comes to character positions/offsets.  Having a space injected causes all of the subsequent tagged text to be off by one, and Bad Things happen (for us).  It wasn't clear to me why this normalization was necessary; my best guess is that it's not so much needed as a nice-to-have: in _some_ cases, it may make date matching work better... if someone mistypes `12March`, it can get normalized to `12 March` and then recognized as a date.  (I don't _**know**_ that this is why the number normalization happens, but I do know that event without the normalization, we didn't see any adverse date tagging results.)  For us, "fixing" typos like this isn't as critical as keeping the text as-entered.

The only way I could come up with to provide a hook to set any kind of plugin-level configuration was to graft it on as a property of the main export.  If the config hook is not used, the old behavior should be identical to what it was
before.  If, however, compromise is extended as:

```javascript
const nlp = require('compromise');
const numbers = require('compromise-numbers');
const dates = require('compromise-dates');

nlp
  .extend(numbers)
  .extend(dates.config({ normalizeBeforeTagging: false }));
//             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^----- this is new!
```

... then compromise-dates will **not** normalize (numbers) before tagging.

(You can ignore the build files; `plugins/dates/src/index.js` and `plugins/dates/src/01-tagger/index.js` have the relevant changes.  If you like this, but want it without the build files, let me know.  Also let me know if you'd rather have the PR targeting the `dev` branch.)


spencermountain:
Hey Jared, had a busy week and this slipped. let me look at this properly tomorrow
Thank you!

spencermountain:
Hey - so I agree with you big time - .dates() is a get method and should never mutate the original document.
I think this is a big no-no, and I didn't realize it was happening in .dates() but i did know that [clone was leaking](https://github.com/spencermountain/compromise/issues/787).
I have a [huge branch open](https://github.com/spencermountain/compromise/tree/new_cache) with a fix for this, and many other problems, as a v14 version. it has clearer thinking about mutating the document, and should fix this issue properly - 
can you give a specific example of the input whitespace getting borked, so I can make sure?
v14 should land in 2 weeks or so, but I can do a pre-release if you fancy taking a look
cheers

spencermountain:
hey @JaredReisinger the `.dates()` method no-longer mutates the doc in [v14](https://github.com/spencermountain/compromise/releases/tag/14.0.0). This was brutal to fix, but i think you'll find you have a solution now, that won't require passing in any more options.
let me know how it goes,
cheers

PR #922, opened 2022-05-26
Fix some clauses

jfemia:
While I was debugging some of my post-update test failures I noticed that `clauses()` failed to split a string like "two whole ducks, necks and giblets reserved".

I wanted to try and fix it myself before bugging you with another bug report, so I added a reproducer test ("an apple, washed and dried") and a bunch of other tests for the other statements you had in `clauses.js`, and then fixed the failing ones. 

I don't know if you had other strings in mind for the expressions in question - but as they were written, they didn't succeed with the strings that the comments indicated they should support - if there's other cases that will fail due to this change we can add them to the test and figure out a suitable match expression


spencermountain:
this is great! Thank you!!

spencermountain:
will release as 14.2.0 on monday.

Issue #918, opened 2022-05-13
missing Plugin type


yesean:
Hi, I'm migrating to compromise v14, so I've replaced `compromise-ngrams` with `compromise-stats`, but I'm using typescript and can't get it to build. I'm getting this error:
```
$ tsc
node_modules/compromise-stats/index.d.ts:4:30 - error TS2724: 'nlp' has no exported member named 'Plugin'. Did you mean 'plugin'?

4 declare const nlpNgrams: nlp.Plugin<
                               ~~~~~~


Found 1 error in node_modules/compromise-stats/index.d.ts:4

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It seems there used to be a [`Plugin` type](https://github.com/spencermountain/compromise/blob/a49cfae80d294f224bd08e2560faf348b9935cc4/builds/types.d.ts#L126) in the `nlp` namespace a while back, but it got removed in the latest version?

spencermountain:
hey Sean, yeah sorry about this. v14 required a re-write of the typefile, and I haven't figured out how to support the plugins properly yet - are you good at this stuff? I'd love a hand
cheers

jfemia:
Isn't the main problem with plugins and types that there needs to be a way to extend the `View` interface itself for plugins that want to add new methods? e.g. so I can do this:
```
import stats from 'compromise-stats'
nlp.plugin(stats)
nlp('something').ngrams({...}) // ngrams is a stats function, but typescript doesn't know it's now a method on View
```

In which case it seems to me as a not-particularly-expert Typescript generics user that the plugin typings should export method interfaces separately, and the `nlp` method typing be given a generic parameter to allow the specified interfaces to be merged with `View`.

I did a little experiment with the type definitions and came up with this method of using plugins:
```
import stats, { StatsMethods } from 'compromise-stats'
nlp.plugin(stats)
const doc = nlp<StatsMethods>('text') // returns type Three & StatsMethods
doc.ngrams({}) // type of ngrams() is correct
```

If you had multiple plugins you'd just construct your doc with all the plugin types you wanted
```
const doc = nlp<SomePluginMethods & SomeOtherPluginMethods & etc...>('text')
```

Does that sound like a reasonable API? I've put my small typing changes in a branch in this commit here - let me know if you want a PR or have a different direction in mind https://github.com/jfemia/compromise/commit/411f0e6889410cdb2994f3418b66d82817d59522

PR #923, opened 2022-05-31
#918 - plugin api typing & stats plugin types update

jfemia:
#918 This doesn't touch anything other than the `d.ts` files so I suppose won't affect plain javascript users, more testing is welcome though :). The default type of `{}` on the `nlp` method also allows existing usage of `nlp()` without any specific plugin types to continue to work.

I didn't look too deep into what the second parameter was for (the empty `{}`) on the original typing, for now I just removed it because it was empty.

### DIFF
--- a/plugins/stats/index.d.ts
+++ b/plugins/stats/index.d.ts
@@ -18,6 +18,6 @@ export interface StatsMethods {
 }
 
 /**  **/
-declare const nlpNgrams: nlp.Plugin<StatsMethods,{}>
+declare const nlpNgrams: nlp.TypedPlugin<StatsMethods,{}>
 
 export default nlpNgrams


jfemia:
"Wasn't me, officer!" - looks like head revision of dev has the same 13 test failures too :)

spencermountain:
hey!! i'm gonna take this for a spin.
thank you so much. i needed this.

spencermountain:
hey Jamie, some plugins extend `nlp` which is working(!), and others extend `View`, which I've not had any luck with -
```ts
export interface DatesMethods {
  dates(): View // <- isn't exported
  times(): View
  durations(): View
}

/** extend compromise View **/
nlp.View<DatesMethods>.  // <-i know nothing..
``` 
 - got any advice?
thanks

spencermountain:
ah, you're right. 

yeah, ideally three types of plugins would be supported:
```js
nlp.workerPool() // speed-plugin

let arr = nlp('foo').ngrams() //on View ✅

let view = nlp('foo').dates() //also returning a View?
```
technically the .dates() example returns a extended View object, with extra methods. I'm ok ignoring that, if it's too weird right now.
Thanks for your help with this. I can't make heads-or-tails of this stuff.

jfemia:
OK so the shorter answer first - a couple of ideas for returning a `View` from the plugin methods:

1) The "quick" way - either import `View` from `compromise/types/view/one` or grab its type directly from the `nlp` object already imported like so:
```
import nlp from 'compromise'
+type View = ReturnType<typeof nlp>
...
```

2) A possible slight usability improvement for the API might be to merge the `View` with the `DatesMethods` so the views returned can be chained with more date-related methods:
```
type View = ReturnType<typeof nlp> & DatesMethods
...
```

---------

More generally on extending module types and wishing to avoid a bunch of generics everywhere, I thought that, in theory, module augmentation would be the way to go here - although I haven't yet been able to find the right secret sauce to getting this working. 

In my mind it should be possible for a plugin to add its stuff straight onto the core interfaces - something like (for example these two sets of plugin methods bundled into one while I was testing...)

```
// this thing would "augment" the module with these methods
declare module 'compromise' {
  export function workerPool(): any // I don't know what this returns but you get the point :)

  export interface View {
    /** match all date-phrases */
    dates(): View
    /** match time-of-day phrases */
    times(): View
    /** match lengths of time, like '2 weeks' */
    durations(): View
  }
}
```

but either it doesn't work quite like this or I'm doing something wrong because my app using these packages doesn't seem to merge these into the main types like i'd expect... this part is pushing the limits of my TS knowledge

PR #942, opened 2022-08-01
fix: type in `compromise-stats`

Himself65:
https://github.com/spencermountain/compromise/blob/98282ff35cfd43bc25ea5085ece90687bd46cc7f/plugins/stats/src/ngram/getGrams.js#L33-L46


spencermountain:
thank you!

PR #950, opened 2022-08-25
Add common honorifics and remove mister from abbreviations

GrumpyLittleTed:
I found that in the sentence "Miss Jones likes cake." the "Miss" honorific was not identified. I traced this back to .../lexicon/people/honorifics.js.  and abbreviations/honorifics.js. 

Basically 'mister' was treated as an abbreviation when it's not so I moved it to people/honorifics
I added miss, missus, lord, lady, baron and baroness to people/honorifics

After an npm run pack and npm run build the changes mean that 'Miss' is now recognised as an honorific.

I noticed that 'judge' is commented out without explanation...but I didn't mess with it...


spencermountain:
hey thanks!
I may take a look at handling some of these ambiguous words more carefully - ('miss' as verb) ('lady' and 'lord' as nouns).
that's a big improvement, nice work!
will add this to the next release, cheers

PR #951, opened 2022-08-26
Add common honorifics and remove mister from abbreviations

GrumpyLittleTed:
Fixes:
In "A certain Miss Jones eats cake." the word "Miss" is not identified as an honorific because it is missing from the lexicon so added.
Mister is defined as an abbreviation but it should be an honorific: moved from abbreviations  to honorifics 
Baron, Baroness and Lord are also missing honorifics: added

Suggestions:
Took me a while to figure out I had to pack before doing build so that I could do my tests - a build_all script in package.json may help newbies like myself.

### DIFF
--- a/data/lexicon/people/honorifics.js
+++ b/data/lexicon/people/honorifics.js
@@ -2,6 +2,8 @@
 export default [
   'admiral',
   'ayatullah',
+  'baron',
+  'baroness',
   'brigadier',
   'captain',
   'chancellor',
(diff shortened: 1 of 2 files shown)


spencermountain:
hey, yeah sorry abt that - this honorific stuff is a mess - i'm trying to clean this up in the dev branch. I've been thinking it doesn't make much sense to tag something as honorific, until it's infront of a #Person tag. "the king" isn't an honorific ... right?
and also- like you found, it's partly duplicated in the sentence tokenizer.  oh boy.

good idea- i'll add a readme file to the data folder.
i'm gonna try and release something for this today, but if i don't get it, you're free to checkout the dev branch and play around
cheers

thegoatherder:
@spencermountain FWIW we use `#Honorific` quite a lot in our implementation. I agree that it doesn't make sense unless it's in front of `#Person`, however, `#Person` isn't always reliable, especially with some ethnic names.

spencermountain:
this has been cleaned-up considerably in 14.5.0 - take if for a spin and let me know
cheers

PR #955, opened 2022-09-07
Add more conjunctions

GrumpyLittleTed:
Add a broader range of conjunctions. Adding some new conjunctions caused some tests to fail due to subsequent level 2 or 3 processing: those conjunctions have been left in conjunctions.js but commented out.


spencermountain:
hey nice!
thank you. Some of these, like 'even' may not always pop-out as conjugunctions. Will add this to the next release.
cheers

PR #968, opened 2022-10-19
ci: update github actions

Fdawgs:
This PR bumps GitHub Actions to their latest major versions, which now use Node 16 internally instead of Node 12 (EOL as of end of April 2022, no longer receives security updates):

- Releases for `actions/cache` can be [found here](https://github.com/actions/cache/releases)
- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)


spencermountain:
hey! i was wondering why that had stopped working.
thank you.

PR #969, opened 2022-10-19
build(deps-dev): remove @rollup/plugin-alias

Fdawgs:
`@rollup/plugin-alias` dev dependency is not used in the codebase or in any scripts.


spencermountain:
good catch! thank you.

PR #990, opened 2023-01-16
Fix types for the dates plugin

evenfrost:
This fixes missing TypeScript types for the `compromise-dates` plugin.


spencermountain:
yikes! thank you!

Issue #973, opened 2022-10-25
Inconsistent tagging of #TextValue,  #Date and #Duration


thegoatherder:
In below similar text examples, the POS tags for `six` is often `#Value` and `#TextValue`, _but not always_
and `month(s)` is `#Date` and `#Duration` _but not always_.

```javascript
const nlp = require('compromise')

const text = [
  'a six month session',
  'I have agreed a six month session',
  'six months',
  `A six-month session`,
  'I have agreed a six-month session',
]

text.forEach((t) => nlp(t).debug())
```

Outputs:

```ascii
  ┌─────────
  │ 'a'        - Determiner
  │ 'six'      - Value, TextValue, Cardinal
  │ 'month'    - Date, Noun, Duration
  │ 'session'  - Noun, Singular



  ┌─────────
  │ 'six'      - Value, TextValue, Cardinal
  │ 'months'   - Noun, Plural



  ┌─────────
  │ 'A'        - Determiner
  │ 'six'      - Value, TextValue, Cardinal
  │ 'month'    - Date, Noun, Duration
  │ 'session'  - Noun, Singular



  ┌─────────
  │ 'I'        - Noun, Pronoun
  │ 'have'     - Verb, Auxiliary
  │ 'agreed'   - Verb, PastTense
  │ 'a'        - Determiner
  │ 'six'      - Noun, Singular
  │ 'month'    - Date, Noun, Duration
  │ 'session'  - Noun, Singular




  ┌─────────
  │ 'six'      - Value, TextValue, Cardinal
  │ 'months'   - Noun, Plural



  ┌─────────
  │ 'A'        - Determiner
  │ 'six'      - Value, TextValue, Cardinal
  │ 'month'    - Date, Noun, Duration
  │ 'session'  - Noun, Singular



  ┌─────────
  │ 'I'        - Noun, Pronoun
  │ 'have'     - Verb, Auxiliary
  │ 'agreed'   - Verb, PastTense
  │ 'a'        - Determiner
  │ 'six'      - Noun, Singular
  │ 'month'    - Date, Noun, Duration
  │ 'session'  - Noun, Singular
```

It seems like these should have a more consistent tagging output?
I can take a look at improving this in a PR if you give me some pointers on where to configure?

thegoatherder:
It's worth pointing out that in Compromise `v13.11.4` the output is different - all occurrences of `six` are recognised as `#TextValue` and `#Value` - so this appears to be a regression in `v14`

```ascii
=====
[...]

PR #1003, opened 2023-03-29
fix(#973): plural #Duration tagging e.g. months

thegoatherder:
Adds `#Duration` tags for plural versions of current durations.


spencermountain:
thanks

PR #1023, opened 2023-07-17
chore: align package.json with ESM module

louisgv:
This is needed to import this package via some strict resolver that validate ESM package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "name": "compromise",
   "description": "modest natural language processing",
   "version": "14.9.0",
-  "main": "./src/three.js",
+  "module": "./src/three.js",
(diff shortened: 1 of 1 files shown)


spencermountain:
hey thanks! I'm working on a release in the next day or two, and will include it in there.

spencermountain:
hey - do you think it'd be a problem for me to use both `main` and `module`?
hoping to avoid doing a major release. hard to keep up with this spec
cheers

louisgv:
@spencermountain hey sorry for the late reply - I think it should be ok! The old CJS resolver only care for the `main` field and would ignore other :-?

spencermountain:
released as `14.10.0` - thanks!

PR #1030, opened 2023-09-02
Added University Rule :hammer:

MarketingPip:
This PR adds a simple rule -

```js
{ match: 'university of #Place', tag: 'Organization', reason: 'university-of-Foo' }
```

Not sure how to add this as a place as well without adding another duplicate line.

<br>

ps; another good rule might be -

```js
{ match: '#Place university', tag: 'Organization', reason: 'university-of-Foo' }
```

### DIFF
--- a/src/2-two/postTagger/model/nouns/places.js
+++ b/src/2-two/postTagger/model/nouns/places.js
@@ -16,6 +16,8 @@ export default [
   { match: '(district|region|province|municipality|territory|burough|state) of #ProperNoun', tag: 'Region', reason: 'district-of-Foo' },
   // in Foo California
   { match: 'in [#ProperNoun] #Place', group: 0, tag: 'Place', reason: 'propernoun-place' },
+  //University of Foo
+  { match: 'university of #Place', tag: 'Organization', reason: 'university-of-Foo' }
   // Address 
   { match: '#Value #Noun (st|street|rd|road|crescent|cr|way|tr|terrace|avenue|ave)', tag: 'Address', reason: 'address-st' },
 ]


spencermountain:
Good idea! Thank you

MarketingPip:
@spencermountain - I messed up! I put this into the wrong file! (should have been in orgs) - we will move it - or you can feel free to! Another one is "#Place beach" - for example ```fort myers beach```.

Apologizes for blowing up your notifications lately too but you'll be more than happy with what I have been working on lol!

spencermountain:
Yeah, please at least run the tests before making a pull-request. It's not the best github etiquette

MarketingPip:
@spencermountain - woof, glad you got /caught that "," misssing. 🤦 Sorry for the more than rookie mistake. Won't happen again. That said - when I add rules etc, something like this in future (ps - I will move this over to right place with some new rules to follow). 

And can you give me a detailed description of how to add a test for rule sets? Like this one added?

spencermountain:
of course!
there's always a gold-star for any PR with a test that currently fails, but now passes. That's brain-candy for me, but also prevents the next person from forgetting about it.

Tests can go anywhere. There's a bunch of tagging tests [here](https://github.com/spencermountain/compromise/blob/master/tests/two/match.test.js#L1140) that could work. If you wanted to add a test for `fort myers beach` there, that would be awesome.
if you add it, then run `npm run test` . You can also run just one file, using [tape](https://www.npmjs.com/package/tape) if you'd like.

So if i remember, there is a custom organization-specific tagger thing [here](https://github.com/spencermountain/compromise/blob/master/src/2-two/preTagger/compute/tagger/3rd-pass/03-orgWords.js) for things like 'society of X' and stuff. I think it was simply faster than running the loop inside the match() function. You can see the words [here](https://github.com/spencermountain/compromise/blob/master/src/2-two/preTagger/model/orgWords.js), which are a bit of a mess. 
cheers

PR #1052, opened 2023-09-17
Added interjections 🔨

MarketingPip:
Added a ton of missing expressions / interjections. All have been found from a frequency list of 30,000 most commonly used words & tagged via Wiktionary. Build tests have been passed previously before submitting this PR.  :hammer: 

> ps; notice that 30,000 k mark - I have made some other lists to again based off this. Some dude named Spencer I was video / lecture was talking about this being average mark too for human vocabulary. 😉 So figured this would be a good idea / rule to build lexicon on.


spencermountain:
hey Jared - I think there may be some misunderstandings with adding words to the lexicon. 
'yummy' is an adjective - 'it was yummy' - 'boy' and 'man' are very common nouns.
Adding these words to the lexicon will change every occurance of 'man' to an Interjection.

If you add these words, and run `npm run pack`, you'll see the tests do not pass.
Let me know if I can help
cheers

MarketingPip:
@spencermountain funny you brought that up! Been at stacking verb lists up for us etc and seen / forget they were separated by various types 😩. But yeah - awesome point - my stupid ass just assumed since the build and tests passed this would be good.

If you want to help clean this list (would be appreciated). And if have a solution that can even help sort this + the other verbs I have collected for compromise etc! Feel free to share! As I think we both prefer not sorting 3000 + verbs 😩 lol

MarketingPip:
@spencermountain - awesome job! And thank you for your work / dedication to Compromise!

PR #1063, opened 2023-11-21
Add safety for reserved words in buildNet

VitaliyR:
Hi!

I was playing around with my pet project - and suddenly added word `constructor` as matcher for `buildNet`:

```
const matches = [
  { match: 'constructor' }
]
const net = nlp.buildNet(matches)
```

And when tried to run the code - it crashed because it tries to call `push` method on `object.constructor` which is a function.

I've fixed it via explicitly checking for types. What do you think?

P.S. Also not sure if thats the best place for test - feel free to point me into right direction. Thanks!


spencermountain:
This is beautiful Vitalii!
thank you!

PR #1077, opened 2024-01-05
Html fix

### LINKED ISSUE #1073: If there is a Pre tag, the text from that term is not included in the HTML export
https://github.com/spencermountain/compromise/blob/4ef66b3e5798c63f3f0f3b7935ffae1597b6dd3b/src/1-one/output/api/html.js#L67

Test case: 
```javascript
  doc = nlp(`if i can recall, my grey dog loves pizza crusts (they are really good).`)
  html = doc.html({ '.red': 'my grey dog', '.blue': doc.match('loves') })
  t.equal(
    html,
    `if i can recall, <span class="red">my grey dog</span> <span class="blue">loves</span> pizza crusts (they are really good).`,
    here + 'html two classes'
  )
```

Currently returning:
```html
if i can recall, <span class="red">my grey dog</span> <span class="redTwo">loves</span> pizza crusts ( are really good).
```

suggestion is:
```javascript
out += t.pre || ''
out += t.text || ''
```

Any issues with that?

ryancasburn-KAI:
Fixes #1073 and adds a test case.

### DIFF
--- a/src/1-one/output/api/html.js
+++ b/src/1-one/output/api/html.js
@@ -1,12 +1,12 @@
 const isClass = /^\../
 const isId = /^#./
 
-const escapeXml = (str) => {
+const escapeXml = str => {
   str = str.replace(/&/g, '&amp;')
   str = str.replace(/</g, '&lt;')
   str = str.replace(/>/g, '&gt;')
   str = str.replace(/"/g, '&quot;')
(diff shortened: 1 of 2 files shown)


spencermountain:
thank you!
will release this, and anything else we get up to, next week
🏅

ryancasburn-KAI:
Great thanks!

Issue #1080, opened 2024-01-18
Using .freeze() in nlp.plugin()? 


thegoatherder:
The documentation for [`.freeze()`](https://observablehq.com/@spencermountain/compromise-freeze) shows adding a word to the lexicon and freezing it using `addWords()`.

For projects that use the [`nlp.plugin()`](https://observablehq.com/@spencermountain/compromise-plugins) option to initialise the lexicon at program start, an object is passed to the `words` property with a name-value pair syntax for each tag.

Is there a way to specify these tags to be frozen from within the plugin object?

thegoatherder:
Can it also be specified in a match-object?

```javascript
{
  match: "#Diagnostic+ (preassessment|assessment|reassessment|pre-assessment|re-assessment)",
  reason: "",
  tag: "Diagnostic",
}
```

spencermountain:
Hey Adam, good idea - I’ll look at adding both of these features over the
next few days.
Let me know if there’s anything else you can find
Cheers

thegoatherder:
Thanks, Spence! Once these are implemented we can fully integrate it into our solution and get it tested thoroughly. We will let you know if we find any anomalies! Looking forward to getting this feature rolling…!

spencermountain:
hey Adam, both features are implemented on dev, and i will test, document, and release it, hopefully tomorrow.
cheers

thegoatherder:
Hey Spence that’s great! I forgot to mention the reason we need the match-object is for buildNet() and sweep() - will these support frozen tags too?

spencermountain:
yep - `{match:'foo', tag:'Bar', freeze:true, }` will lock in #Bar, and it will be unchangeable. Will that do the trick for your purpose?

thegoatherder:
Yep absolutely perfect, thank you!

spencermountain:
both features have been released in `14.2.0`:

frozen lexicon via plugin:
```js
nlp.plugin({
  // normal lexicon
  words:{
    foo:'Bar'
  },
  // frozen lexicon
  frozen: {
    'juicy fruit': 'Singular',
    'front steps': 'Plural',
  },
})
let doc = nlp(`i ate juicy fruit on the front steps`)
doc.debug()
```

and `freeze` inside sweep:
```js
let matches = [
  { match: 'juicy fruit', tag: 'Singular', freeze: true },
  { match: 'front steps', tag: 'Plural', freeze: true },
]
[...]

PR #1083, opened 2024-01-26
chore(misc): update types with new freeze related properties

Fdawgs:
As [mentioned here](https://github.com/spencermountain/compromise/issues/1080#issuecomment-1907044784), `Match` has a new `freeze` property and `Plugin` has a new `Frozen` property.


spencermountain:
Thank you!

PR #1084, opened 2024-01-26
ci: bump github actions

Fdawgs:
This PR bumps GitHub Actions to their latest available versions.
- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)

This PR also removes `actions/cache` and uses the caching options from `actions/setup-node`, which functions the same.


spencermountain:
👌

PR #1087, opened 2024-02-06
ci(release): update workflow

Fdawgs:
This PR:

-   Removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials` to false. They are not used after the initial checkout, and this stops them from accidentally leaking through a script
-   Declares the minimum permissions for the workflows to run at the job level, following principle of least privilege; see [related GitHub security post](https://securitylab.github.com/research/github-actions-building-blocks/)
-   Removes the `cache-dependency-path` param from the `actions/setup-node` action, as if `cache` is declared then it will use the default path anyway
-   Updates `npm publish` script to [publish with provenance](https://docs.npmjs.com/generating-provenance-statements#publishing-packages-with-provenance-via-github-actions) to provide assurance that published package was generated from this repo

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,9 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
 
     # Note that these steps are *identical* to build-and-test (with the caveat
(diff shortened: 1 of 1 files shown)


spencermountain:
wow cool! Thank you so much.
I didn't know about any of that stuff. Nice one!

PR #1093, opened 2024-02-29
ci: update github actions

Fdawgs:
Same as https://github.com/spencermountain/compromise/pull/1087, but applied to the test workflows.


prisis:
Maybe you have interest to look into https://www.stepsecurity.io

PR #1104, opened 2024-04-25
chore: export view types

Fdawgs:
I was previously able to declare a variable's type in VS Code using the following:
```js
/** @type {import('compromise/types/view/three').default} */`
const test = nlp.view;

console.log(test.sentences());
```

This would allow Intellisense to kick in, so if i hovered over `test.sentences()` i'd get the following:
image

However it seems the latest minor version of TypeScript which shipped with VS Code has removed that.
This PR updates package.json to explicitly export the view types, so that the type can continue to be declared like so:
```js
/** @type {import('compromise/view/three').default} */
const test = nlp.view
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,15 @@
       "import": "./src/three.js",
       "require": "./builds/three/compromise-three.cjs",
       "types": "./types/three.d.ts"
+    },
+    "./view/one": {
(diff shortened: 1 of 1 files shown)


spencermountain:
thank you!
sorry, not sure what's happened with the github script.
I owe ya a drink, Frazer.

PR #1106, opened 2024-04-29
chore(package): export misc types

Fdawgs:
Forgot to include this in #1104, the `Net` type in the misc types is handy!


Fdawgs:
Tests are failing because `mac-os-latest` is now using ARM.
See https://github.com/fastify/.github/issues/37

PR #1137, opened 2024-08-07
feat(plugins/dates): support duration ranges

### LINKED ISSUE #1135: Compromise-dates plugin: Support duration ranges
At present, the compromise-dates plugin supports `date-ranges` such as `aug 1 - aug 31`.

It would be great if it also supported duration ranges and project those accordingly, for example:
- `8 - 10 weeks`
- `1 to 2 days`

As an example:
```js
'use strict'

/** @type {import('compromise').default} */
const nlp = require('compromise')
const nlpDates = require('compromise-dates')
nlp.plugin(nlpDates)

const text = 'I will see him in 8-10 weeks'

const doc = nlp(text)

const context = {
  today: '2023-01-01',
}

const datesViews = doc.dates(context)
const dateJson = datesViews.get()[0]

console.log(dateJson)
// Would expect this to output:
// {
//   start: '2023-02-26T00:00:00.000Z',
//   end: '2023-03-12T23:59:59.999Z',
//   timezone: 'Europe/London',
//   duration: { years: 0, months: 0, days: 14, hours: 0, minutes: 0 }
// }
```

I'm happy to have a crack at this myself but not sure where to start?

Fdawgs:
Closes #1135 

This will need a squash PR as it's bound to be a mess.


plugins/dates/src/api/parse/range/02-date-range.js
```diff
@@ -138,4 +139,25 @@ export default [
...
+      let start = new Unit(context.today, null, context)
+      let end = start.clone()
+
+      const duration = unit.text('implicit')
+      start = start.applyShift({ [duration]: min.numbers().get()[0] })
+      end = end.applyShift({ [duration]: max.numbers().get()[0] })
+
+      return {
+        start: start,
+        end: end,
```
Fdawgs, line 159: Need to sort a conditional where if the duration phrase is `day`, `month`, or `year` then subtract a day and call `end.end()`. 

```js
      end = end.applyShift({ [duration]: max.numbers().get()[0], day: -1 })

      return {
        start: start,
        end: end.end(),
```

spencermountain: oh, right - hmm.
There's different types of `Unit` there if you wanted to use them - the idea was that they each call `.end()` or `.middle()` and it means something different for a Month or an Hour. Not sure it worked out v. well, of if that helps in your case.
Worse you could grab the spacetime object with `Unit.d` (sorry!) and add a condition there
cheers

Fdawgs: Thanks @spencermountain, I'll give it a try but may just target bigger units then like day/week/month/year for now. The plugin is called `compromise-dates` after all, not `compromise-times`.

Should i throw tests for this in a existing file (if so which?), or do you think this needs a new one?

spencermountain: sure, up to you!

plugins/dates/src/api/parse/range/02-date-range.js
```diff
@@ -138,4 +139,30 @@ export default [
...
+      let end = start.clone()
+
+      const duration = unit.text('implicit')
+      start = start.applyShift({ [duration]: min.numbers().get()[0] })
+      end = end.applyShift({ [duration]: max.numbers().get()[0] })
+
+      // Ensure that the end date is inclusive
+      if (!['day', 'days'].includes(duration)) {
+        end = end.applyShift({ day: -1 }).applyShift({ [duration]: 1 })
+      }
```
Fdawgs, line 160: Hopefully this is an okay way of doing this.
I found that day and days was projecting inclusively, but, other durations weren't.

For example `1-2 weeks` was returned as 8 days rather than 14 days.

spencermountain: sounds good!

Fdawgs: Great, thanks @spencermountain!
Hopefully last question, do i need to do anything extra for handling past tense in this?
I.e. `2 to 3 weeks ago`.

spencermountain: oh, no problem - i would make a new section, in addition to this '2 to 4 weeks' one, with similar logic.
go for it!

oh,  one thing - maybe add a `'^in'` on the match, so it doesn't interfere with `2 to 3 weeks after june 3rd` etc.
lookin good! thanks for the help.

Fdawgs: Cheers @spencermountain, have added the `^in` as requested for the future tense match block.

Struggling with the past tense, as you can see from the now failing tests! When you get a spare second could you take a look at the past test block please? Thought it'd be as simple as flipping the logic.

spencermountain: sure - i can merge this to dev and fix the failing tests, if you're ready?

spencermountain: hey Frazer, haven't forgotten about this -  been really busy, then had covid. Will try to get to it shortly.
cheers

Fdawgs: No worries!

Fdawgs: Thanks for taking this on @spencermountain, https://github.com/spencermountain/compromise/commit/42f9cd0b6ee40401ad89fb375fd2207de3757a5d appears to have stopped the ranges from being inclusive?

PR #1146, opened 2024-09-21
Fix issues with `replaceWith`

tony-scio:
1. The `keep` options are now respected when using the function-param
   variant of `replaceWith`.
2. `keep: case` now respects the case of the original target, instead of
   basing the decision on the first term of a sentence. I believe this is
   the expected behavior.
3. Unmatched replacements no longer crash when used with keep options.
   Demonstrated by: `replace-with-no-match`.
4. `pre` text is now preserved. Demonstrated by: `replace-with-phone-number`
5. Add type definitions for `keep`, there's some outdated documentation
   around that refers to old names for these params. Also, `possessives`
   was undocumented. This makes it easier to use correctly.
6. Add a test for `keep: possessives`.


spencermountain:
ah! thank you Tony!
This is some real complicated stuff in there, and this is a great PR.
Let me put it on master, and kick the tires a bit before the next release.
🏅 oof, even the tests and types file too, oof. 🏅 
thank you

tony-scio:
Thanks for the quick merge! Feel free to let me know if you notice issues w/ it.

PR #1165, opened 2025-01-15
fix: TypeScript resolution should now work across bundlers

### LINKED ISSUE #1164: TypeScript Support Impacted By Incorrect Package.json Configuration
Hey there! First of all; thank you immensely for this package - your work on it is awesome and deeply appreciated. 😄 

While working on integrating this package in a project, I ran into problems with importing the packages, especially the subpackages. It's a little difficult to pin-down between `import` and `require` and what bundlers, but when I took a look at `AreTheTypesWrong` (a utility that helps maintainers like us figure out how different bundlers resolve TypeScript types), I noticed things were off:

image

https://arethetypeswrong.github.io/?p=compromise%4014.14.3

In comparison, this is what a correctly configured library `package.json` looks like to TypeScript via dual-CJS/ESM publising:

image

This may be the root cause of some previous issues like:

https://github.com/spencermountain/compromise/issues/1012

https://github.com/spencermountain/compromise/issues/845

https://github.com/spencermountain/compromise/issues/862

And more.

I'll be opening a PR to address this problem without modifying your bundling solution. I'm also trying to dogfood this PR in my project to confirm it fixes the issues we're running into.

Thanks!

crutchcorn:
This PR fixes #1164 by:

- Fixing fallback cases
- Adding `node10` (think Webpack 4 and older node) resolution via `{root}/one/package.json` files
- Creating CJS-specific `.d.cts` types so TypeScript doesn't incorrectly warn of `default` usage when doing a `require`
- Removing unused (AFAICT) type exports like `./misc` and `./views`

This should not be a breaking change for existing users, but will likely fix **many** reported problems in CJS (and even some ESM) usage of this library

# Before

Using "AreTheTypesWrong" CLI:

image

# After

Using the same CLI:

image

### DIFF
--- a/one/package.json
+++ b/one/package.json
@@ -1,0 +1,25 @@
+{
+	"name": "compromise-one",
+	"version": "14.14.3",
+	"description": "",
+	"type": "module",
+	"module": "./../src/one.js",
+	"main": "./../src/one.js",
+	"types": "./../types/one.d.ts",
+	"exports": {
+		"./package.json": "./package.json",
+		".": {
+			"import": {
+				"types": "./../types/one/one.d.ts",
+				"default": "./../src/one.js"
+			},
+			"require": {
+				"types": "./../types/one.d.cts",
+				"default": "./../builds/one/compromise-one.cjs"
+			}
+		}
+	},
+	"author": "Spencer Kelly <spencermountain@gmail.com> (http://spencermounta.in)",
+	"license": "MIT",
+	"sideEffects": true
+}
--- a/package.json
+++ b/package.json
@@ -11,29 +11,29 @@
   "types": "./types/three.d.ts",
   "exports": {
     ".": {
+      "types": "./types/three.d.ts",
       "import": "./src/three.js",
-      "require": "./builds/three/compromise-three.cjs",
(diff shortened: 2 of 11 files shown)


spencermountain:
whoaaaaaa
This is amazing. Thank you Corbin.
Yes, i believe your suspicions are correct. I read something in 2015 about esm and haven't changed anything since then. Thank you for showing me this tool.

Gonna merge this and will see if I can get a release going.

spencermountain:
okay, one question - do we need the `one/package.json` `two/package.json` etc?

I kinda like the `compromise/one` `compromise/two` scheme ... and would prefer not having to publish individual packages `compromise-one` etc?

are those package.json files scooped up by a bundler somewhere?
thank you for your help. I know none of this.
cheers

crutchcorn:
(total aside; thanks so much for being so bright and kind in these replies ✨  Always such a refreshing welcome to any new project one contributes to :D )

spencermountain:
ok released as `14.14.4` - I'm around tomorrow, so please let me know if you see anything weird.
Thank you for your help.

PR #1174, opened 2025-02-26
build: ensure generated umd files are strict

### LINKED ISSUE #1173: Rollup creating non-strict cjs and umd files
Node version: 20.18.1
Compromise version: 14.14.4

The `.cjs` and `.js` files created by Rollup are missing the `"use strict";` directive that enables strict mode. The ESM files created don't need it as they are strict by default.

Strict mode can improve performance by eliminating some JavaScript features that hinder optimizations. It also helps avoid subtle bugs by enforcing more consistent scoping and variable declarations.

The [MDN article on strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) alludes to it, but the V8 JS engine used by Node will use more optimised execution paths when strict mode is enabled. See [related Stack Overflow discussion](https://stackoverflow.com/questions/38411552/why-use-strict-improves-performance-10x-in-this-example) for an example.

It looks to be the `@rollup/plugin-terser` options doing this; Rollup converts the output to cjs/umd then `@rollup/plugin-terser` is ran over the results. As it is configured to `module: true` it is stripping out the `"use strict"` directive and other useful cjs/umd bits.

This is occurring in all Rollup configs in this repo.

**Suggested fix:**

https://github.com/spencermountain/compromise/blob/d2eed9bb6fac36700d24b1f50986f5ff0666a7d9/rollup.config.js#L5
Create a shallow copy of the above config and pass that to terser for the cjs/umd outputs.

```js
const opts = { keep_classnames: true, module: true }
const umdOpts = { ...opts, module: false }

// ...

plugins: [nodeResolve(), terser(umdOpts), sizeCheck({ expect: 69, warn: 15 })],
```

Also change the output format for the `.cjs` files from `umd` to `cjs` to remove amd and iife cruft.

Happy to make a PR to resolve this!

Fdawgs:
closes #1173 

Also updated `rollup` and `@rollup/plugin-node-resolve` to their latest versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "compromise",
-  "version": "14.14.3",
+  "version": "14.14.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "compromise",
-      "version": "14.14.3",
+      "version": "14.14.4",
       "license": "MIT",
       "dependencies": {
         "efrt": "2.7.0",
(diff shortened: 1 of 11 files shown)


spencermountain:
thanks!

PR #1179, opened 2025-04-22
style: use `const` for never reassigned variables

### LINKED ISSUE #1178: Use of `let` for variable declarations?
Whilst sifting through the code, I noticed that near all of the variable declarations use `let` even in cases where the variables are not reassigned.
Is there a particular reason for using `let` over `const` in such cases or is it just a stylistic choice?

Would you be opposed to changing these unassigned variables to `const` to improve readability?
Happy to make a PR.

Fdawgs:
closes #1178. Just toggled on `prefer-const` rule in eslint and ran eslint over code.


spencermountain:
gonna trust you on this one!
thanks

PR #1187, opened 2025-07-16
Improve typescript cjs support for dates

noseworthy:
The `compromise-dates` plugin was building a commonjs package, but the types weren't properly exported in a way that TypeScript was able to import them.

By providing an `index.d.cts` file and updating the `exports` property of the `package.json` for this plugin we're able to now properly import `compromise-dates` in a commonjs TypeScript project.

### DIFF
--- a/plugins/dates/index.d.cts
+++ b/plugins/dates/index.d.cts
@@ -1,0 +1,3 @@
+import dates from './index.d';
+
+export = dates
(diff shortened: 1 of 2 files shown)


spencermountain:
oh man, this stuff never ends!
Thanks Michael. Will add this to the upcoming release.
cheers

PR #1197, opened 2026-02-23
Switch from declare module to declare namespace for tsgo support

orta:
This PR allows folks using the nightly builds of tsgo to be able to use the lib

The term 'modules' in .d.ts files is pretty old (it had it's [hey-day in TS 1.5](https://www.typescriptlang.org/docs/handbook/namespaces.html)) and now the new version of TypeScript built in go (7) is dropping support for the old `module` -> `namespace`  alias which was baked into the compiler


spencermountain:
agh - thanks!

PR #1213, opened 2026-07-15
fix(match): include last word in capture groups spanning wildcards

### LINKED ISSUE #1139: Last word missing from capture groups
Node version: 20.16.0
Compromise version: 14.14.0
Compromise-dates version: 3.6.0

Bit of an odd one, when using capture groups with the `.` syntax, the last word is always missing:

```js
'use strict'

/** @type {import('compromise').default} */
const nlp = require('compromise')
const nlpDates = require('compromise-dates')
nlp.plugin(nlpDates)

const text = 'I ate some cheeseburgers with Mr. Lahey'

const doc = nlp(text)
let result = doc.match('[#PastTense .{0,10} cheeseburgers]', 0)
console.log(result.out('text'))
// Outputs 'ate some', expected 'ate some cheeseburgers'

result = doc.match('[#PastTense some cheeseburgers]', 0)
console.log(result.out('text'))
// Outputs 'ate some cheeseburgers' as expected

result = doc.match('[#PastTense .{0,10} cheeseburgers]').groups(0)
console.log(result.out('text'))
// Outputs 'ate some', expected 'ate some cheeseburgers'
```

xianjianlf2:
Fixes #1139

## Symptom

A named capture group that wraps a greedy `.`/`.*` wildcard together with surrounding tokens drops the tokens after (and mis-sizes the span around) the wildcard:

```js
nlp('one two three after').match('[<all>one . after]').group('all').text()
// actual:   'one two three'
// expected: 'one two three after'
```

Reproduced by @spencermountain in the issue thread.

## Root cause

In `src/1-one/match/methods/match/steps/astrix.js`, `doAstrix` sets the group result with:

```js
g.length = skipto - state.t
```

The assignment overwrites whatever length earlier regs in the same group had already contributed, so the group is re-anchored to just the wildcard span. The tokens captured before the wildcard are dropped from the group, and the trailing literal ends up outside the reported range.

## Fix

Accumulate instead of overwrite:

```js
g.length += skipto - state.t
```

Leading tokens stay in the group, and the trailing literal is appended by the normal match loop as before.

## Tests

Added `tests/two/match/greedy-capture.test.js` covering the issue's repro plus variants (leading literal + wildcard, wildcard + trailing literal, group starting at the wildcard, and greedy min/max behavior). Full suite passes: 13,419 assertions, 0 failures (`npm test`).

### DIFF
--- a/src/1-one/match/methods/match/steps/astrix.js
+++ b/src/1-one/match/methods/match/steps/astrix.js
@@ -25,7 +25,9 @@ const doAstrix = function (state) {
   // set the group result
   if (state.hasGroup === true) {
     const g = getGroup(state, state.t)
-    g.length = skipto - state.t
+    // accumulate onto any tokens already captured before the wildcard,
+    // so '[one .* after]' keeps its leading (and trailing) tokens
+    g.length += skipto - state.t
   }
   state.t = skipto
   // log(`✓ |greedy|`)
--- a/tests/two/match/greedy-capture.test.js
+++ b/tests/two/match/greedy-capture.test.js
@@ -43,6 +43,32 @@ test('issue-654: greedy capture', function (t) {
   t.end()
 })
 
+/*
+ * a capture group spanning a wildcard must keep the tokens
+ * on both sides of the '.*'
+ * https://github.com/spencermountain/compromise/issues/1139
+ */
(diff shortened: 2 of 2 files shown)


spencermountain:
wow, thank you!
this is very helpful!!

PR #1224, opened 2026-09-17
Recognize more openers of a secondary clause in mainClause

dualfroz:
mainClause() missed a few ways a sentence can open with a secondary clause, so the parser took the the verb from that clause instead of the main. Anyway, just as you asked @spencermountain i tried to help a little more. on a 20 sentence probe of main clause openers, i managed to get it from 14 to 18/20.


spencermountain:
So good! Thank you!!

PR #1225, opened 2026-09-17
Skip the verb of a relative clause following the subject

dualfroz:
a relative clause right after the subject had its verb taken as the verb of the whole sentence. `The boy who you saw at the store committed a robbery.` gave `saw`. uncommented two of the cases already sitting in svo.test.js


spencermountain:
Haha, amazing. Lemme fix this in the morning.
Thank you for your help.

